Issue #8, opened 2017-08-22
Refactor for getting devices by HA type


upsert:
As per comments in home-assistant/home-assistant#8903, the Home Assistant authors have suggested having a method to retrieve devices by their HA type such as 'light', 'switch' and 'cover'. The refactor is needed before support for Triathlon shades can be added.

For reference, find below all the device types supported on the LEAP interface. Includes a sensor type in anticipation of one day supporting Pico remotes which will become sensors.

 'light'
---------- |
 WallDimmer
 PlugInDimmer

 'switch'
---------- |
WallSwitch

'cover'
-------- |
SerenaHoneycombShade
SerenaRollerShade
TriathlonHoneycombShade
TriathlonRollerShade
QsWirelessShade

'sensor'
-------- |
Pico1Button
Pico2Button
Pico2ButtonRaiseLower
Pico3Button
Pico3ButtonRaiseLower
Pico4Button
Pico4ButtonScene
Pico4ButtonZone
Pico4Button2Group
FourGroupRemote

Other Device Types
------------------ |
SmartBridge

@gurumitts I hope you don't mind if I add a few issues and if I can learn a bit more Python, hopefully I'll contribute some code too. I'll also try to write up what I've learned about the interfaces.

gurumitts:
No issue at all.  =)

gurumitts:
Changes are now live in version 0.2.8.  https://pypi.python.org/pypi/pylutron_caseta/0.2.8

PR #10, opened 2017-08-30
Get devices by domain

upsert:
Adding code for issue #8.

First commit, f381de2, adds docstrings for functions, cleans up PEP8 and other minor issues found during development. I've attempted to follow [style guidelines](https://home-assistant.io/developers/development_guidelines/) also used in Home Assistant and to be consistent with spelling, line lengths, use of double-quotes for sentences and other minor style issues. Processed with pylint and flake8.

Second commit 4b7154a is the new method to call from Home Assistant to retrieve devices by their domain such as 'light', 'switch' etc. I've added 'sensor' even though it is not currently used from Home Assistant code. The original methods remain intact and with docstrings added to clarify their intended use.

My first attempt at Python so hopefully I didn't break anything and it makes sense!

### DIFF
--- a/pylutron_caseta/smartbridge.py
+++ b/pylutron_caseta/smartbridge.py
@@ -1,4 +1,4 @@
-"""Provides an API to interact with the Lutron Caseta Smartbridge."""
+"""Provides an API to interact with the Lutron Caseta Smart Bridge."""
 
 import json
 import logging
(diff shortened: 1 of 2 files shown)


pylutron_caseta/smartbridge.py
```diff
@@ -42,60 +41,116 @@ def __init__(self, hostname=None):
...
+        :param type_: LEAP device type, e.g. WallSwitch
+        """
         devs = []
         for device_id in self.devices:
-            if self.devices[device_id]['type'] in _types:
+            if self.devices[device_id]['type'] == type_:
+                devs.append(self.devices[device_id])
+        return devs
+
+    def get_devices_by_types(self, types):
```
gurumitts: We should mark this is deprecated in favor of the get_by_domain().  We can remove this in a future version.

upsert: @gurumitts Do you consider this as a standalone Caseta library? The function could be used for various reasons even if it is not from Home Assistant. For example, I could get all the 3 button Picos if I know what I'm looking for.

That said, if the primary use of the library is assumed to be Home Assistant (and indeed the 'light', 'cover', 'switch', etc. comes from HA) and you want to, I can mark it deprecated. How does one do that in Python? Do I add a logger warning statement? Let me know which you prefer.

gurumitts: Ahh... good point.  Lets leave it alone for now.  Yes it is meant to me a stand alone lib.  I am merging this now.

gurumitts:
@upsert Your changes are now live in version 0.2.8 in pypi.  You can use pip to install it.

PR #15, opened 2017-10-14
Update to use SSL interface

thinkelastic:
Hi,

As you may know Lutron has closed SSH access to the caseta hub. Somebody has figured out how to authenticate with the hub and use the SSL just as the phone applications do.

This is the script to authenticate:
https://gist.github.com/mhofman/f8e1fdd5dce49abacf5fd776fb3727ce

For more context see here:
https://community.home-assistant.io/t/lutron-caseta-leap-ssh-integration/29136/20

This pull request updates your library to use SSL to communicate with the hub instead of SSH. Once you merge it and update pypi I will work on update home assistant.


pylutron_caseta/__init__.py
```diff
@@ -3,35 +3,6 @@
 # This is the private key for the Lutron Smart Bridge 'leap' user.
 # https://github.com/njschwartz/Lutron-Smart-Pi/blob/master/RaspberryPi/LutronPi.py
```
jobi: that comment should probably be removed too

thinkelastic: Makes sense :-)

pylutron_caseta/smartbridge.py
```diff
@@ -39,6 +39,8 @@ def __init__(self, hostname=None):
         monitor.start()
         for _id in self.devices:
             self.get_value(_id)
+
+        self.set_value('41', 0)
```
jobi: I'm a complete newcomer to this but I think this line could definitely use some comment

thinkelastic:
Unfortunately the oauth part requires a UI interaction (web) that cannot be done through the library. That is why I propose we do it through the home assistant UI. I believe other components do the same.

I share your fustration, thankfully it seems we might have a way to fix it.

nriley:
This code does not seem to work with Python 3.6 for me (perhaps it was developed on Python 2?).  I replaced the Unicode string literals with byte literals but even after doing so, I still get a hang on 'loading devices':

```
(PyLMS3) [1100] ~/Documents/Development/PyLMS3/src/pylutron-caseta%python3 -c "import pylutron_caseta.smartbridge; pylutron_caseta.smartbridge.Smartbridge('lutron-[...].local.', 'lutron.key', 'lutron.appcert', 'lutron.remotecert')"
Connecting to Smart Bridge via SSL
Successfully connected to Smart Bridge.
Loading devices
^CTraceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/nicholas/Documents/Development/PyLMS3/src/pylutron-caseta/pylutron_caseta/smartbridge.py", line 37, in __init__
    self._load_devices()
  File "/Users/nicholas/Documents/Development/PyLMS3/src/pylutron-caseta/pylutron_caseta/smartbridge.py", line 289, in _load_devices
    ssl_output = self._ssl_sock.recv(1)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/ssl.py", line 987, in recv
    return self.read(buflen)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/ssl.py", line 865, in read
    return self._sslobj.read(len, buffer)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/ssl.py", line 627, in read
    v = self._sslobj.read(len)
KeyboardInterrupt
```

(after waiting several minutes, I interrupted it).  Is it working for anyone else?

mdonoughe:
I had the same problem as @nriley and was able to fix get farther by checking for b'\n' instead of '\n' to find the end of the line, but my lights still aren't turning on.

https://github.com/mdonoughe/pylutron-caseta/commit/1e8291274cd1eb41c3cef60473cd425f05480c75

There has to be a better way of reading a stream of JSON objects coming over a socket in Python.

kfcook:
@mdonoughe your code works fine for me (I spent an hour fiddling with code I downloaded last night...wish I would have checked for updates sooner!). I'm able to turn lights and switches on and off.

mdonoughe:
You're right, but my Home Assistant was not happy. I made a couple more changes and now things seem to be working again for me with this code https://github.com/mdonoughe/pylutron-caseta/tree/6ea865fcee684cff2398d45f49e5c9595bc279e7 .

kfcook:
I’ll have to pull down your code and test shortly. With the existing code, I can only keep the socket up for about 15 minutes. After that, it closes and doesn’t reestablish until a reload of HA.

mdonoughe:
I thought I had it working because I didn't get startup errors and I was able to turn the lights on and off a few times, but then I saved everything and tried to turn on the light and it didn't work. It looks like the problem is that `ssl.SSLError` is being thrown, not `ConnectionError`, so `logged_in` does not get set back to false. I'm trying with `ssl.SSLError` now, but it seems like logged_in should be set to false regardless of the specific exception thrown. I'll update my repo if it still works later.

nriley:
Just confirming that @mdonoughe's changes do at least make connections to my Smart Bridge work for me.  I'm not using it with home-assistant currently so even if a persistent HTTP connection won't be reliable, it's better than nothing!  Thanks.

gurumitts:
I will test this and merge it soon.  Hopefully tonight or tomorrow evening.

kfcook:
I wouldn’t merge it yet. It works for a few minutes but then throws an exception and crashes. I haven’t had a chance to troubleshoot yet, but others are working on it.

gurumitts:
Understood... I will also take a look into that issue.  Thanks all for the contributions!

mdonoughe:
I put some new code on my branch late last night and this morning not only was home-assistant still working but the lights came on, although it seems like maybe it's slow to respond if you catch it when the connection is dropped but the client hasn't noticed yet.

I ended up rewriting the networking using asyncio. I have suspiciously few error handlers but somehow it seems to work anyway and without hanging threads when the connection resets. I don't do a lot of python and I'm pretty sure the code I've written only works in 3.6 even if asyncio was available before then. I don't know if that's a problem. For me running home-assistant from the docker image I already have 3.6. Maybe it's as easy as replacing `async def` with coroutine generators or whatever. I haven't tried anything older.

Locally I have some additional changes from yesterday where I was trying to figure out the testing and linting. There weren't any unit tests to speak of when I started and the version of mypy it was using didn't like 3.6 syntax, and upgrading mypy to the latest is giving me virtualenv errors.

I don't know anything about this LEAP protocol that I didn't get from reading this code and I haven't played with the timeouts. Do you think it would be possible to keep the connection alive by sending blank lines or dummy commands every ten minutes?

nriley:
This takes quite a while to connect for me — longer than the previous non-async version:
```
10/16 09:27:11 Connecting to Smart Bridge via SSL
10/16 09:27:24 Successfully connected to Smart Bridge.
10/16 09:27:24 Loading devices
10/16 09:27:24 sending b'{"CommuniqueType": "ReadRequest", "Header": {"Url": "/device"}}'
```
Also it'd be nice to figure out why you have to bypass certificate checking…

kfcook:
@mdonoughe you should just be able to send `{"CommuniqueType":"ReadRequest","Header":{"Url":"/server/1/status/ping"}}` every so often to the hub to keep the connection alive assuming the timeout timer starts from the time the last command was sent and not total connection time.

mdonoughe:
13 seconds to open a connection seems like a really long time. Does that happen consistently for you with the asyncio code I put up last night? On my system it opens the connection and reads all the initial information in the same second. I have a Pentium J2900 and 4GB RAM connected to the bridge by ethernet.

I noticed in @thinkelastic's code that he had certificate checking turned off. I tried turning it on, but it didn't work and I was more interested in getting the code to work than preventing somebody physically in my house from taking control of a light switch so I turned it back off. The bridge actually has two certificates. `get_lutron_cert.sh` gets a `local_signs_app_certificate` which it calls `remote_cert` and I assumed it was the CA certificate of the bridge, but it's not. It looks like `local_signs_app_certificate` is a certificate signed by your bridge but owned by Lutron and used to issue client certificates, which I guess allows you to add/update apps while the bridge is unreachable. This means during the pairing process we should be generating the following files:

1. client.key: `$private_key`
2. client.crt: `$app_cert\n$remote_cert`
3. ca.crt: `openssl s_client -tls1_2 -connect $server_addr:8081 -showcerts -prexit 2>/dev/null | openssl x509 -outform pem`

mhofman:
To get the self signed root CA, you'll need to pull it out of the bridge. The bridge uses it as its SSL cert so something like `echo | openssl s_client -showcerts -connect ${bridgeip}:8081 2>/dev/null | openssl x509 -inform pem` would work.

To make sure it's the right one, you could probably validate the remote cert (or app / remote chain) against it.

nriley:
Yes, it really does take 12-14 seconds consistently. I just updated to the latest version and reorganized my certs/keys as @mdonoughe and @mhofman recommended so now the cert verification passes, but it's still just as slow:

```
pylutron-caseta% ts python3 -c "import pylutron_caseta.smartbridge; pylutron_caseta.smartbridge.Smartbridge('lutron-[...].local.', 'client.key', 'client.crt', 'ca.crt')"
10/16 22:27:08 Connecting to Smart Bridge via SSL
10/16 22:27:23 Successfully connected to Smart Bridge.
```

Connecting via the iOS app or HomeKit (or executing openssl s_client as above) is pretty much instantaneous.

Unfortunately I don't know too much about Python asyncio but when I sampled the python process there were a couple threads waiting on the GIL (I think) and one thread stuck waiting on a kqueue. I tried switching to PollSelector and SelectSelector and while that first thread now waited on something else, it didn't seem to be any faster. So either this is a Python bug or we're doing something wrong...

mdonoughe:
Does it still take 12-14 seconds if you use an ipv4 address instead of the mdns name?

mdonoughe:
I added a loop that does a ping about once a minute and my connection has stayed alive nearly two hours so far.

nriley:
@mdonoughe - bingo!  It was failing to connect over IPv6.  I added `family=socket.AF_INET` to `asyncio.open_connection` and it connected reliably within a second even with the mDNS name specified.

gurumitts:
Working on this a bit.  I am running into a issue that was reported earlier. about the `b'literal\n'`  Is this working for folks without converting to bytes?

> TypeError: a bytes-like object is required, not 'str'

mdonoughe:
@gurumitts I may have confused everything by sort of hijacking this pull request. I've been talking about the code I have in my tls branch that I forked from @thinkelastic's branch for this pull request. That was the first thing I fixed.

I should add the AF_INET change from @nriley, somehow add testing, and I guess figure out if it can be made compatible with Python 3.4 or 3.5.

grahamr:
One note as well - i blocked the hub from hitting firmwareupdates.lutron.com from my pihole (you could use firewall as well) 

This SHOULD keep us in a known state once sorted out.

GervaisdeM:
How's this going? Is there any way I can use this right now with my non-pro hub?

mdonoughe:
@GervaisdeM I've been using my tls branch with my non-pro hub. I have a dockerfile that replaces the caseta code in my home-assistant with the latest code from that branch. I want to get unit tests working, make it possible to disconnect from the hub, and get rid of the background thread when loading into an environment that already has an asyncio event loop (like home-assistant).

scottocs11:
@mdonoughe How can I load this component in Hass.io?

mdonoughe:
I know next to nothing about hass.io.

kfcook:
@mdonoughe just to add a data point, I've also been using your tls branch since Saturday and it's been performing perfectly. No lagging connections and no random disconnects/crashes. Thanks for all of your work on this!

Have you also been working on the HASS side of this? I suppose we'll also need to build something for the UI to generate the certificates/handle all of the OAUTH stuff.

rdbahm:
@scottocs11 @GervaisdeM I blindly stumbled my way through getting it to work with Hassio (And regular Hass) last night as a custom component (overriding default lutron_caseta).

1. Use the script to authenticate that's provided in the top of the pull request. Get the files and save them in your config folder.
2. Save my updated version of lutron_caseta.py to custom_components in your config folder: https://github.com/rdbahm/home-assistant/blob/pylutron-upgrade/homeassistant/components/lutron_caseta.py
3. In your config, in the lutron_caseta component, add lines for "keyfile," "certfile," and "ca_certs." For me, this looks like:
```
lutron_caseta:
  host: 192.168.1.75
  keyfile: 'privatekey.key'
  certfile: 'appcert.crt'
  ca_certs: 'remotecert.crt'
```

Some notes:
- This uses my own version of pylutron_caseta. I'm bad at Git and tend to break stuff even when I don't mean to. Consider this a temporary solution.
- My fork of pylutron_caseta is a fork of @mdonoughe 's work. I needed to change it for my purposes because certificate verification wasn't working and because I couldn't get the dependency to load in HASS without a tag.
- So far, it's working very smoothly.
- I think we'll need to do some refinement on the authentication script to make this more user-friendly eventually.

Sorry for the pull request hijack, all. I wanted to be able to test and develop on this in HASS and didn't find any other complete guide. Thanks to @thinkelastic and @mdonoughe especially for your work.

scottocs11:
@rdbahm Thanks so much!

geoffp:
@mdonoughe, if you'd be willing to share your Dockerfile, I'd be happy to add another data point.

mdonoughe:
@geoffp it's in https://github.com/mdonoughe/home/tree/master/image .

grahamr:
@rdbahm another data point for your notes - this worked for me in hass.io

geoffp:
I've got it mostly figured out, but I'm not strong enough with SSL certs to figure out how to get the output of the authentication script plugged into the `lutron_caseta` config. I'll have to pick this up tomorrow.

gurumitts:
I have ping implemented on a test branch.

mdonoughe:
@geoffp The script produces only 3/4 of the things you need. The two certificates are your client's certificate chain, the key is your client's key, but you also need the CA certificate from the bridge itself. You can get it from `openssl s_client -connect <bridge>:8080 -showcerts`.

@gurumitts I have had ping working on my branch. The version from this pull request also had some issues with status updates after reconnection, which shouldn't happen as often with pinging, but is still a possibility.

I want to finish adding tests and get it hooked into home-assistant as async rather than launching a new event loop on a background thread.

gurumitts:
@mdonoughe I fixed the status notification errors after reconnect in my branch.  I am going to hold off on merging this.  I looked at your branch and I like the changes.  I will push the reconnect fixes into your branch.  Please send a PR when you are happy.

mdonoughe:
@gurumitts I think I'm sort of good now. I have green checkmarks on Travis CI (I had to change the configuration) and it seems to be working on my Home Assistant. I'll write up a description and send you a pull request tomorrow.

gurumitts:
Awesome work @mdonoughe !!

PR #16, opened 2017-11-01
Switch from SSH to TLS to support new firmware

mdonoughe:
Since a few weeks ago when Lutron released new firmware they removed our ability to control the bridge devices over SSH. @thinkelastic swapped out the SSH for TLS connections and opened gurumitts/pylutron-caseta#15, but that code was written for Python 2, and simply switching out SSH for TLS doesn't seem to be enough because the the bridge connections would expire after about fifteen minutes, and it seems like there may have been an older issue in pylutron-caseta which caused sync issues after reconnecting to the bridge (I was able to control the lights, but the status updates weren't appearing in Home Assistant). I've been using versions of this code for a couple weeks now and it seems reliable.

I think I've fixed everything and then fixed some things that weren't really broken and then fixed the CI which seems to have been broken by changes in the machines used by Travis CI.

1. SSH is now TLS. paramiko is now asyncio.
2. When reconnecting, the status of all devices is required the same as on the first connection.
3. The bridge is sent a ping command about once a minute to keep the connection alive.
4. The code for reading and writing the basic LEAP protocol (CRLF-delimited JSON objects) has been split out into a separate file. A connection factory is passed to the constructor, but a class method exists to generate the typical connection factory and pass it to the constructor for you.
5. The code to make the initial connection has been removed from the constructor, and there is now a method to disconnect. These are exposed as asyncio coroutines. A background thread is no longer spawned to own the connection. This means the caller is responsible for providing an asyncio event loop, either by being natively asyncio or by launching a thread to host the asyncio event loop. If the caller only needs to open the connection, send a command, and disconnect, it should be possible to use `event_loop.run_until_complete(bridge.connect()); bridge.turn_on('1'); event_loop.run_until_complete(bridge.close())` without setting up a full asyncio environment.
6. Unit tests have been introduced.
7. The CI files have been updated based on the ones from home-assistant/home-assistant. This includes suppressing warnings-as-errors about methods taking too many parameters and types having too many instance attributes, pinning pytest-asyncio to the last version to support Python 3.4. Additionally, the CI now includes Python 3.6.

I don't do a lot of Python programming so I might have done some things weird or not quite right.

Possible issues:

- If the bridge closes the connection and refuses new connections, or if the OS on the host machine does the same, the program may immediately retry after every refusal. This shouldn't be too bad because each connection request should result in control yielded to other tasks running in the event loop so we shouldn't actually impact the application much besides spamming a lot of SYN packets to the bridge until it comes back. Maybe there should be some sort of backoff delay.
- Looking at the unit test output, the ping task doesn't seem to always cancel immediately when the connection is closed. Maybe the ping task needs to be saved as another instance variable and explicitly cancelled.
- When reconnecting, subscribers probably receive an update even though the value has not actually changed. This is probably nothing, but this could be suppressed by comparing the incoming state against the current state before propagating the message from the bridge to subscribers.


gurumitts:
@mdonoughe do you have a pr for HA as well?

PR #20, opened 2018-03-07
Use fully qualified names for devices

### LINKED ISSUE #18: Device name uses incomplete name.
https://github.com/gurumitts/pylutron-caseta/blob/52dfa7000b7cd23bee586671d8ef839c932178b8/pylutron_caseta/smartbridge.py#L321

Device name is assigned the `Name` value from the devices dict, however, the `FullyQualifiedName` would be more appropriate as the latest Lutron Caseta app enforces the use of Rooms. The room name is the prefix the `FullyQualifiedName` and left out of `Name`. This leads to multiple switches/dimmers named `Light` or `Lights` vs `Kitchen Lights`, `Kitchen Table Lights`.

Changing line 321 to `device_name = " ".join(device["FullyQualifiedName"])` seems more appropriate.

samdoran:
With enough lights, it gets very easy to have a name collision. In Home Assistant, the device names change each time HASS is restarted, causing unreliable control of devices. Using the fully qualified name fixes this issue by naming devices in a more global fashion.

Fixes #18

### DIFF
--- a/pylutron_caseta/smartbridge.py
+++ b/pylutron_caseta/smartbridge.py
@@ -318,7 +318,7 @@ class Smartbridge:
             if 'LocalZones' in device:
                 device_zone = device['LocalZones'][0]['href']
                 device_zone = device_zone[device_zone.rfind('/') + 1:]
-            device_name = device['Name']
+            device_name = '_'.join(device['FullyQualifiedName'])
             device_type = device['DeviceType']
             self.devices[device_id] = {'device_id': device_id,
                                        'name': device_name,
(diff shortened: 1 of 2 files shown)


rdbahm:
This seems like a solid change that will make the name used by Home Assistant much more sensible and consistent for the end user. It'll be a breaking change for existing users, unfortunately, but I think that's worth it.

samdoran:
I wasn’t sure the best way to handle the breaking change. It is definitely an improvement, though. I’ve been using this patch for a couple months and it alleviated all the inconsistencies I was having controlling my Caseta lights from Home Assistant.

Is there a good way to handle the breaking change other than mentioning it in the release notes in Home Assistant?

gurumitts:
Version 0.4.0 contains this change:  

https://pypi.python.org/pypi?:action=display&name=pylutron_caseta&version=0.4.0

PR #25, opened 2018-04-21
add model and serial to device listing

mdonoughe:
In order to support the [Home Assistant Entity Registry](https://www.home-assistant.io/docs/configuration/entity-registry/) we need to be able to construct a stable identifier value for every device. I don't know the details of how Lutron issues serial numbers, but I assume a combination of model ID and serial number will be unique within every home.


mdonoughe:
Related: home-assistant/home-assistant#14013

mhofman:
Wondering, isn't the device ID already stable ?

mdonoughe:
There are probably a few ways it could change, such as if you got a new bridge or did a factory reset or maybe even some firmware updates. The model and serial number of the switches themselves should be more reliable.

PR #30, opened 2018-07-18
improve reconnection logic

### LINKED ISSUE #19: reconnection failures
I probably introduced this error somehow. I've noticed that occasionally my home automation will lose its connection to the bridge and I'll get messages like this in my logs:

```
2018-01-09 07:24:27 DEBUG (MainThread) [pylutron_caseta.smartbridge] Connecting to Smart Bridge via SSL
2018-01-09 07:24:57 ERROR (MainThread) [homeassistant.core] Error doing job: Task exception was never retrieved
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/asyncio/tasks.py", line 180, in _step
    result = coro.send(None)
  File "/usr/local/lib/python3.6/site-packages/pylutron_caseta/smartbridge.py", line 301, in _ping
    self._writer.write({
AttributeError: 'NoneType' object has no attribute 'write'
```

followed by:

```
2018-01-09 12:53:28 INFO (MainThread) [homeassistant.core] Bus:Handling <Event call_service[L]: domain=light, service=turn_on, service_data=entity_id=light.upstairs_hub_lights, service_call_id=139889797154184-1056>
2018-01-09 12:53:28 ERROR (MainThread) [homeassistant.core] Error doing job: Task exception was never retrieved
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/asyncio/tasks.py", line 180, in _step
    result = coro.send(None)
  File "/usr/src/app/homeassistant/core.py", line 1031, in _event_to_service_call
    yield from service_handler.func(service_call)
  File "/usr/src/app/homeassistant/components/light/__init__.py", line 271, in async_handle_light_service
    yield from light.async_turn_on(**params)
  File "/usr/local/lib/python3.6/asyncio/coroutines.py", line 210, in coro
    res = func(*args, **kw)
  File "/usr/src/app/homeassistant/components/light/lutron_caseta.py", line 57, in async_turn_on
    to_lutron_level(brightness))
  File "/usr/local/lib/python3.6/site-packages/pylutron_caseta/smartbridge.py", line 182, in set_value
    return self._writer.write(cmd)
AttributeError: 'NoneType' object has no attribute 'write'
```

every time I try to control the lights until I restart the process.

It happens infrequently enough that I think it could be triggered by Lutron restarting the bridge by pushing out firmware updates or something like that. It looks like maybe the new connection times out?

mdonoughe:
- try harder to handle connection errors
- terminate the connection if the bridge stops answering pings

Fixes #19

### DIFF
--- a/pylutron_caseta/leap.py
+++ b/pylutron_caseta/leap.py
@@ -60,9 +60,9 @@ class LeapWriter(object):
         """Initialize the writer."""
         self._writer = writer
 
-    def close(self):
-        """Close the underlying stream."""
-        self._writer.close()
+    def abort(self):
+        """Abort the underlying stream."""
+        self._writer.transport.abort()
 
     @asyncio.coroutine
     def drain(self):
--- a/pylutron_caseta/smartbridge.py
+++ b/pylutron_caseta/smartbridge.py
@@ -12,6 +12,10 @@ _LOG = logging.getLogger(__name__)
 _LOG.setLevel(logging.DEBUG)
 
 LEAP_PORT = 8081
+PING_INTERVAL = 60.0
+PING_DELAY = 5.0
+CONNECT_TIMEOUT = 5.0
+RECONNECT_DELAY = 2.0
 
 
 class Smartbridge:
@@ -33,6 +37,8 @@ class Smartbridge:
         self._reader = None
         self._writer = None
         self._monitor_task = None
+        self._ping_task = None
+        self._got_ping = asyncio.Event(loop=self._loop)
 
     @asyncio.coroutine
     def connect(self):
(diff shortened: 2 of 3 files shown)


mdonoughe:
Okay I think this is really ready now. At first it looked like it would work based on the tests I added, but actually it did not work because:

1. I was calling StreamWriter methods which I had forgotten were not present on LeapWriter.
2. Sometimes the exception raised when failing to connect to the bridge is OSError, not ConnectionError.
3. After reconnecting, the device dictionary would contain all new devices, and any external code that was holding onto instances of the original devices would no longer see updated values. (in Home Assistant this resulted in write-only devices)

The new behavior should be as follows:

Every 60 seconds, SmartBridge sends a ping to the physical bridge like before, but now if the bridge does not respond within 5 seconds the connection is terminated. The connection will also be terminated if the process of connecting to the bridge and getting the initial status information takes more than 5 seconds. If the connection is closed or terminated by either SmartBridge or the physical bridge, the connection will be retried after 2 seconds.

The monitor loop now logs fatal errors that cause it to exit.

I have this running on my server (technically the home branch on my fork) and it has survived a few instances of me intentionally interrupting the power to my bridge.

I'm not that happy about the logging that occurs during reconnection. It's probably sufficient just to log the type and details of exception without the traceback.

PR #39, opened 2019-11-19
Prepare for 0.5.1 release

mdonoughe:
- Device model and serial information is exposed
- Less likely to encounter certificate errors
- Better reconnection after network errors
- No longer fails when fan controllers are present
- Basic support for fan controllers

All that's left is to increase this version number, merge to master, tag, and push to PyPI?


mdonoughe:
I think I did it.

PR #40, opened 2020-02-23
Do not process scenes attached to engraved keypads

swails:
New engraved picos (released with RadioRA2 Select but compatible with Caseta) operate by creating virtual buttons/scenes that they trigger (this allows them to control different sets of lights with different buttons, rather than requiring each button in a pico to control the same set of lights).

Those scenes do not have a `Name` attribute, causing scene detection to crash pylutron-caseta for any system that has these devices.  This will skip assignment of those.

### DIFF
--- a/pylutron_caseta/smartbridge.py
+++ b/pylutron_caseta/smartbridge.py
@@ -407,7 +407,7 @@ class Smartbridge:
                 break
         for scene in scene_json['Body']['VirtualButtons']:
             _LOG.debug(scene)
-            if scene['IsProgrammed']:
+            if scene['IsProgrammed'] and 'Name' in scene:
                 scene_id = scene['href'][scene['href'].rfind('/') + 1:]
                 scene_name = scene['Name']
                 self.scenes[scene_id] = {'scene_id': scene_id,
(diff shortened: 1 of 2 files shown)


swails:
Examples of these picos that, when installed in a Caseta system, starts spitting out name-less scenes:

* [here](https://www.mrsupply.com/lutron-pj2-2b-gwh-p01-pico-2-button-entry-scene-wireless-remote-white.html?gclid=Cj0KCQiA4sjyBRC5ARIsAEHsELE41cr0nCvuLi4QnIEjylbnzWp8oebF0yUe_iib5TGEOdl-N2fdfmIaAn2yEALw_wcB)
* [here](https://www.mrsupply.com/lutron-pj2-4b-gwh-p03-pico-4-button-any-room-scene-wireless-remote-white.html?gclid=Cj0KCQiA4sjyBRC5ARIsAEHsELGEf5gwtn0RyqQAs605HjjjRXAnDWO1KmNUUBcPKStW7cPfovVgfDcaAn4uEALw_wcB)

mdonoughe:
Can you put a couple comments or something in the code so in the future its obvious why these vbuttons are special and are being ignored?

swails:
I'll do that in #41 when I add support for the fancy new Caseta occupancy/vacancy sensors.

swails:
Closed in favor of #41

swails:
Comment added.

Issue #27, opened 2018-06-30
get_lutron_cert.py remote_signs_app_certificate error


hdemel:
I'm unable to successfully run this script. I always get the following error:

```
Traceback (most recent call last):
  File "get_lutron_cert.py", line 78, in <module>
    with open('caseta.crt', 'rb') as f:
FileNotFoundError: [Errno 2] No such file or directory: 'caseta.crt'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "get_lutron_cert.py", line 125, in <module>
    app_cert = pairing_response["remote_signs_app_certificate"]
KeyError: 'remote_signs_app_certificate'

```

I've tried running this on two separate Ubuntu Servers (18.04). Instructions in the script were followed exactly.  I've been able to run this before but as of today this doesn't work.

Any advice on how I can get this working?

mdonoughe:
Are you using the script from this repo or the script on the Home Assistant documentation? The one on Home Assistant needs to be updated next time the Home Assistant integration is updated.

mbrancato:
I get the same thing. When I enable debug with requests, I get the error below. It looks like the path `/api/v1/remotepairing/application/user` is not valid.

```
Open Browser and login at https://device-login.lutron.com/oauth/authorize?client_id=***************************************************************&redirect_uri=https%3A%2F%2Fdevice-login.lutron.com%2Flutron_app_oauth_redirect&response_type=code
Enter the URL (of the "error" page you got redirected to (or the code in the URL): ***************************************************************
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): device-login.lutron.com
send: b'POST /oauth/token HTTP/1.1\r\nHost: device-login.lutron.com\r\nUser-Agent: python-requests/2.18.4\r\nAccept-Encoding: gzip, deflate\r\nAccept: */*\r\nConnection: keep-alive\r\nContent-Length: 332\r\nContent-Type: application/x-www-form-urlencoded\r\n\r\n'
send: b'code=***************************************************************&client_id=***************************************************************&client_secret=***************************************************************&redirect_uri=https%3A%2F%2Fdevice-login.lutron.com%2Flutron_app_oauth_redirect&grant_type=authorization_code'
reply: 'HTTP/1.1 200 OK\r\n'
header: Server: Cowboy
header: Date: Fri, 13 Dec 2019 12:29:21 GMT
header: Connection: keep-alive
header: Cache-Control: private, no-store
[...]

PR #44, opened 2020-03-21
rewrite get_lutron_cert

mdonoughe:
This version of get_lutron_cert does not depend on Lutron web services (which have disappeared), or even having a Lutron cloud account at all. You pair with the bridge device by pressing the button on the back while the script is running.

### DIFF
--- a/get_lutron_cert.py
+++ b/get_lutron_cert.py
@@ -172,10 +172,14 @@ ssl_context.use_privatekey(LAP_KEY)
 
 
 class JsonSocket:
+    """A socket that reads and writes json objects."""
+
     def __init__(self, socket):
+        """Create a JsonSocket wrapping the provided socket."""
         self._socket = socket
 
     def read_json(self):
+        """Read an object."""
         buffer = b""
         while not buffer.endswith(b"\r\n"):
             buffer += self._socket.read(1024)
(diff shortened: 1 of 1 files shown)


swails:
the button flow is better than going through device-login.lutron.com, but that web service isn't gone... what were you referring to?

mdonoughe:
It's not? People were complaining it didn't work anymore and it didn't work when I last tried it.

mdonoughe:
Fixes #27

swails:
> It's not? People were complaining it didn't work anymore and it didn't work when I last tried it.

I just tried the old version and it worked.

Just to be clear, though, the button approach is better (and is the approach that would be needed to implement a config flow for Home Assistant).

mhofman:
This is neat! Out of curiosity, is this local pairing protocol documented anywhere? How was it discovered? Wondering where that `DisplayName` ends up used, or what other roles exist. Also, where do the "Caseta Application" cert and private key come from?

mdonoughe:
This is from analyzing the Android app. I wrote up some documentation about it but never got around to completing and publishing it anywhere.

mhofman:
Ha my original script came from analysing the Android app too, but I don't believe there was a local pairing feature back then, which is why it relied on the cloud provisioning.

PR #47, opened 2020-03-26
treat OSError like ConnectionError

### LINKED ISSUE #45: Lutron Caseta quits working with Home Assistant after a while
Is there any chance that the recent changes for 0.6.0 will fix the issue documented here: https://github.com/home-assistant/core/issues/31361 ?

Otherwise, is this a problem that should be addressed in this repo or is it a Home Assistant issues?  Any other debugging that I can do, please let me know.

mdonoughe:
Sometimes Python incorrectly reports OSError "No route to host" which does not derive from ConnectionError when you fail to create a network connection. pylutron_caseta should continue attempting to reconnect if it gets such an error.

This just retries after any OSError. It doesn't try to figure out what OS its running on and what that error code means for the current OS.

Fixes #45

### DIFF
--- a/pylutron_caseta/smartbridge.py
+++ b/pylutron_caseta/smartbridge.py
@@ -272,7 +272,10 @@ class Smartbridge:
                     _LOG.debug('received LEAP: %s', received)
                     if received is not None:
                         self._handle_response(received)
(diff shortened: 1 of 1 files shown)


swails:
This has actually happened to me a couple times awhile ago. But I haven't been able to observe it since or implement a fix.

PR #62, opened 2020-10-31
expect tags on occupancy group updates

mdonoughe:
Unlike the unsolicited status updates for lights, switches, etc, the occupancy group status updates must be subscribed to, and they are sent with the same tag value that was used for the original subscription.

Related: home-assistant/core#42620


mdonoughe:
Hey, Travis is back!

PR #64, opened 2020-11-03
Add SerenaTiltOnlyWoodBlind to cover types

sebmck:
This adds support for Serena Wood Blinds that were introduced earlier this year.

### DIFF
--- a/pylutron_caseta/__init__.py
+++ b/pylutron_caseta/__init__.py
@@ -14,6 +14,7 @@ _LEAP_DEVICE_TYPES = {
         "TriathlonHoneycombShade",
         "TriathlonRollerShade",
         "QsWirelessShade",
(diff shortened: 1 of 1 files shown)


sebmck:
Actually, while this does make them show up, trying to change the position results in a bridge error of:

> GoToLevel not supported for zone type: Tilt

Seems like there's more that needs to be done to properly support it.

mdonoughe:
I think you want `GoToTilt` and the value should be passed as `Tilt`.

sebmck:
Won't be able to get around to properly finishing this. I was only using this library via homeassistant to control my wood blinds via homekit (which weirdly aren't supported?). I ended up writing my own plugin for homebridge in JS which I'm more familiar with. Thanks for the great project! It was a lot easier with this reference.

Issue #60, opened 2020-10-31
Latest changes have broken home assistant


swails:
See home-assistant/core#42653 - I can reproduce the problem (HA can receive and update light statuses but can't send any commands to the Lutron Smart Bridge).

I think it boils down to the client tag not being used correctly in this library.  The intent behind the client tag is that each client uses the same tag for every single request.  We should pick something like `PYLUTRON_CASETA` or something recognizable rather than a random GUID each time.  It also means we can't use the client tag in order to recognize responses to specific requests.

Fun fact, some Lutron devices (like connect bridges) actually have a maximum number of distinct clients that can connect due to the nature of the legacy hardware that it interacts with (HWQS Proc and RRA2 Main Repeater), and after a fairly small number of messages actually starts rejecting everything.

swails:
This is going to require a fairly hefty refactor given that the client tag is being used to pair requests and responses in a way that was never intended by the Lutron bridge, so I don't really have time to tackle it.  The response is supposed to be paired to the request by matching the ReadResponse to the ReadRequest on the same href with the same client tag.  If the same client sends two requests really quickly, you can't be sure which response was triggered by which request.

mdonoughe:
Are you sure? I have been using this version for weeks and I have restarted Home Assistant many times because I have been testing other things without encountering any startup issues.

In general, the client tag pattern usually allows the client to pass any arbitrary data so that it can associate returned response messages with the client-side operations. At least last I checked, the Android app works the same way, and sends many different client tag values.

If the user were having the problem you describe where sending multiple simultaneous requests causes the bridge to send multiple responses with the same client tag, the user would be receiving "Was not expecting message with tag" errors in the log.

Also, in any case it does not make sense that it would hang this way. Smartbridge implements a five second timeout for every request to the bridge, including the setup requests, so it should not be possible for setup to hang in this way.

More likely, this has always been broken, just not in a way the user noticed. IIRC in the old version, certain requests sent during initialization did not receive a response. It should be possible to tell by enabling debug mode and analyzing the messages sent and received, but also it might be possible to tell by simply disconnecting the bridge from the network, waiting for the connection timeout to occur, and checking the logged stacks for BridgeDisconnectedException.

[...]

PR #65, opened 2020-11-05
use setdefault for areas and occupancy

mdonoughe:
This should fix a problem @gmcmicken reported on #60 where his occupancy groups stop updating in Home Assistant after restarting the bridge connection.


swails:
LGTM

PR #68, opened 2021-01-14
One touch pairing

bdraco:
Call `pair`, press the button.  Get the assets back.

```
>>> from pylutron_caseta.pairing import pair
>>> pair("192.168.107.116")
{'key': '-----BEGIN PRIVATE KEY-----\n<SNIP>\n-----END PRIVATE KEY-----\n', 'cert': '-----BEGIN CERTIFICATE-----\n<SNIP>\n-----END CERTIFICATE-----\n', 'ca': '-----BEGIN CERTIFICATE-----\n<SNIP>=\n-----END CERTIFICATE-----\n', 'version': 1.113}
```


bdraco:
@mdonoughe If this is acceptable, can a release be made as well?

mdonoughe:
Can you make it work with asyncio?

pylutron_caseta/pairing.py
```diff
@@ -0,0 +1,192 @@
...
+    """A socket that reads and writes json objects."""
+
+    def __init__(self, raw_socket):
+        """Create a JsonSocket wrapping the provided socket."""
+        self._socket = raw_socket
+
+    def read_json(self):
+        """Read an object."""
+        buffer = b""
+        while not buffer.endswith(b"\r\n"):
```
mdonoughe, line 64: Why not use readline? This implementation probably works but is incorrect and may fail. We have a stream-oriented connection to the bridge so we may receive the message terminator and another message in the same read.

pylutron_caseta/pairing.py
```diff
@@ -0,0 +1,192 @@
...
+from cryptography.x509.oid import NameOID  # type: ignore
+
+from .assets import LAP_CA_PEM, LAP_CERT_PEM, LAP_KEY_PEM
+
+LOGGER = logging.getLogger(__name__)
+
+SOCKET_TIMEOUT = 10
+BUTTON_PRESS_TIMEOUT = 180
+CERT_SUBJECT = x509.Name(
+    [x509.NameAttribute(NameOID.COMMON_NAME, "get_lutron_cert.py")]
```
mdonoughe, line 22: We probably want a different subject name here, or even a way for the caller to provide the subject name.

pylutron_caseta/pairing.py
```diff
@@ -0,0 +1,192 @@
...
+                    "Header": {
+                        "RequestType": "Execute",
+                        "Url": "/pair",
+                        "ClientTag": "get-cert",
+                    },
+                    "Body": {
+                        "CommandType": "CSR",
+                        "Parameters": {
+                            "CSR": csr_text,
+                            "DisplayName": "get_lutron_cert.py",
```
mdonoughe, line 111: I'm not sure what this DisplayName is for, but it should probably be pylutron-caseta or settable by the caller.

mdonoughe (changes requested):

bdraco:
For the review comments, this is copied code from get_lutron_cert.py. 

I can refactor it to asyncio but I'm about to be physically away from my Lutron hub for a few weeks so I won't be able to test any changes.

How do you feel about merging it as-is so we can get the config flow into home assistant and doing the refactor as a follow up when I get back home in a few weeks?

pylutron_caseta/pairing.py
bdraco: Adjusted to `pylutron_caseta`

pylutron_caseta/pairing.py
bdraco: Adjusted to `pylutron_caseta`

pylutron_caseta/pairing.py
bdraco: This is copied code from `get_lutron_cert.py`

bdraco: Replaced with asyncio

bdraco:
I was able to get this done and tested before I had to leave. Its asyncio now

mdonoughe (approved):

mdonoughe:
Released as [0.8.0](https://pypi.org/project/pylutron-caseta/0.8.0/).

bdraco:
> Released as [0.8.0](https://pypi.org/project/pylutron-caseta/0.8.0/).

TYVM

PR #69, opened 2021-01-19
Add support for retrieving the LIP device list.

bdraco:
The LIP device list allows pico remotes to be identified
when connecting to devices that also support LIP (pro and ra select). Without
the lip device report the user has to guess which device
id is being reported.

This enabled pico remote support in Home Assistant for Pro and RA Select users via
https://github.com/home-assistant/core/pull/45315 https://github.com/home-assistant/core/commit/e546f6d122a65bb599016c81d8d0ab65f31c1716


tests/test_smartbridge.py
```diff
@@ -180,25 +180,31 @@ async def _accept_connection(self, leap, wait):
         response.set_result(response_from_json_file("devices.json"))
         leap.requests.task_done()
 
-        # Second message should be read request on /virtualbutton
+        # First message should be read request on /server/2/id
```
mdonoughe, line 183: Should this be "Second message"?

tests/test_smartbridge.py
```diff
@@ -435,6 +441,138 @@ async def test_device_list(bridge: Bridge):
     assert devices[0]["device_id"] == "3"
 
 
+@pytest.mark.asyncio
+async def test_lip_device_list(bridge: Bridge):
+    """Test methods getting devices."""
+    devices = bridge.target.get_lip_devices()
+    expected_devices = {
```
mdonoughe, line 448: Can we edit down the test data? It seems like one or two would be enough.

tests/test_smartbridge.py
```diff
@@ -435,6 +441,138 @@ async def test_device_list(bridge: Bridge):
...
+        2: {
+            "Name": "Back Hall Pico",
+            "ID": 2,
+            "Area": {"Name": "Back Hall"},
+            "Buttons": [
+                {"Number": 2},
+                {"Number": 3},
+                {"Number": 4},
+                {"Number": 5},
+                {"Numbaer": 6},
```
mdonoughe, line 458: This "a" is suspicious.

pylutron_caseta/smartbridge.py
```diff
@@ -143,6 +144,10 @@ def get_devices(self) -> Dict[str, dict]:
         """Will return all known devices connected to the Smart Bridge."""
         return self.devices
 
+    def get_lip_devices(self) -> Dict[int, dict]:
```
mdonoughe, line 147: If `lip_devices` is unprefixed, do we also need `get_lip_devices`?

bdraco: Nope, I'll take it out.  It was only there to match `get_devices`

tests/test_smartbridge.py
bdraco: Copy paste error on my part. Will fix

tests/test_smartbridge.py
bdraco: I'll reduce it to two devices.

tests/test_smartbridge.py
bdraco: Yes. will fix

bdraco:
TYVM. Would you mind bumping the pypi package as well?

mdonoughe:
https://pypi.org/project/pylutron-caseta/0.9.0/

bdraco:
> pypi.org/project/pylutron-caseta/0.9.0

TYVM!

PR #81, opened 2021-11-28
add support for pico button events

johninaustin:
This code adds the ability to subscribe to Pico button events using LEAP.

A ReadRequest to /device will have some Pico devices. Each Pico will have a ButtonGroups tag with a buttongroup href and ID.  

A ReadRequest to /button will return a series of Buttons tag linked to the above button group. Each of these is a button on the Pico.  

As an example, a Pico device can have buttongroup 2, and that buttongroup can have buttons 103-107. 

To subscribe to event notifications, send a SubscribeRequest for each button ID. For 103, for example, you’d send the SubscribeRequest with the URL /button/103/status/event

Whenever you press or release that Pico button, you’ll get a ButtonStatus body with a ButtonEvent, and the EventType of “Press” or “Release”.

This code finds any devices with buttongroups, gets the list of buttons, and subscribes to events from those buttons. Some new objects are added to retrieve buttons, subscribe to events from those buttons, and to handle response messages.

### DIFF
--- a/pylutron_caseta/smartbridge.py
+++ b/pylutron_caseta/smartbridge.py
@@ -18,6 +18,7 @@ from . import (
     _LEAP_DEVICE_TYPES,
     FAN_OFF,
     OCCUPANCY_GROUP_UNKNOWN,
+    BUTTON_STATUS_RELEASED,
     BridgeDisconnectedError,
     BridgeResponseError,
 )
@@ -43,6 +44,7 @@ class Smartbridge:
     def __init__(self, connect: Callable[[], LeapProtocol]):
         """Initialize the Smart Bridge."""
         self.devices: Dict[str, dict] = {}
+        self.buttons: Dict[str, dict] = {}
         self.lip_devices: Dict[int, dict] = {}
         self.scenes: Dict[str, dict] = {}
         self.occupancy_groups: Dict[str, dict] = {}
@@ -50,6 +52,7 @@ class Smartbridge:
         self._connect = connect
         self._subscribers: Dict[str, Callable[[], None]] = {}
         self._occupancy_subscribers: Dict[str, Callable[[], None]] = {}
+        self._button_subscribers: Dict[str, Callable[[], None]] = {}
         self._login_task: Optional[asyncio.Task] = None
         # Use future so we can wait before the login starts and
         # don't need to wait for "login" on reconnect.
@@ -140,9 +143,24 @@ class Smartbridge:
         """
         self._occupancy_subscribers[occupancy_group_id] = callback_
 
+    def add_button_subscriber(
+        self, button_id: str, callback_: Callable[[], None]
+    ):
+        """
+        Add a listener to be notified of button state changes.
+
+        :param button_id: button id, e.g., 2
+        :param callback_: callback to invoke
+        """
+        self._button_subscribers[button_id] = callback_
+
     def get_devices(self) -> Dict[str, dict]:
         """Will return all known devices connected to the Smart Bridge."""
         return self.devices
+
+    def get_buttons(self) -> Dict[str, dict]:
+        """Will return all known Pico buttons connected to the Smart Bridge."""
+        return self.buttons
 
     def get_devices_by_domain(self, domain: str) -> List[dict]:
         """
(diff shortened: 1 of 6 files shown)


pylutron_caseta/smartbridge.py
```diff
@@ -140,10 +143,25 @@ def add_occupancy_subscriber(
         """
         self._occupancy_subscribers[occupancy_group_id] = callback_
 
+    def add_button_subscriber(
+        self, button_id: str, callback_: Callable[[], None]
```
mdonoughe, line 147:
```suggestion
        self, button_id: str, callback_: Callable[[str], None]
```
This event handler receives whether the button was pressed or released as a parameter.

pylutron_caseta/smartbridge.py
```diff
@@ -474,6 +492,21 @@ def _handle_one_zone_status(self, response: Response):
...
+
+        if response.Body is None:
+            return
+
+        status = response.Body["ButtonStatus"]
+        button_id = id_from_href(status["Button"]["href"])
+        button_event = status["ButtonEvent"]["EventType"]
+        if button_id in self.buttons:
+            self.buttons[button_id]["current_state"] = button_event
+            # Notify any subscribers of the change to occupancy status
```
mdonoughe, line 506:
```suggestion
            # Notify any subscribers of the change to button status
```

pylutron_caseta/smartbridge.py
```diff
@@ -663,6 +722,21 @@ def _process_occupancy_group(self, occgroup):
             name=f"{self.areas[occgroup_area_id]['name']} Occupancy",
         )
 
+    async def _subscribe_to_button_status(self):
+        """Subscribe to pico button status updates."""
+        _LOG.debug("Subscribing to pico button status updates")
+        self.buttons
```
mdonoughe, line 728: Is this line supposed to be here?

pylutron_caseta/smartbridge.py
```diff
@@ -663,6 +722,21 @@ def _process_occupancy_group(self, occgroup):
...
+        try:
+            for button in self.buttons:
+                response, _ = await self._subscribe(
+                   "/button/"+str(button)+"/status/event", self._handle_button_status
```
mdonoughe, line 732:
```suggestion
                   f"/button/{button}/status/event", self._handle_button_status
```

pylutron_caseta/smartbridge.py
```diff
@@ -663,6 +722,21 @@ def _process_occupancy_group(self, occgroup):
...
+                )
+                _LOG.debug("Subscribed to device status")
```
mdonoughe, line 734: Should this log message be more specific about subscribing to status of a specific button?

Alternatively, if somebody has a lot of pico remotes they would get about 3x as many occurrences of this message in the debug log, right? Maybe this should be outside the loop.

mdonoughe (changes requested):
Some minor comments here, and I notice Black complains that a couple lines are too long. The continuous integration service on this repository doesn't seem to be working anymore, but if you can open the code in Visual Studio Code with the development container option and run `tox` it should give the same results.

Can you update the tests? The existing tests are failing because `Bridge._accept_connection` in `test_smartbridge.py` expects a certain sequence of requests. If you update it to expect requests related to buttons and return some basic button information then the existing tests should start passing again. There should be a test that verifies when a button press event is received the event handler gets called.

pylutron_caseta/smartbridge.py
johninaustin: I'll make this change

pylutron_caseta/smartbridge.py
johninaustin: I'll make this change

pylutron_caseta/smartbridge.py
johninaustin: Nope!  removing...

pylutron_caseta/smartbridge.py
johninaustin: Feels like if you're debugging, having each one display would be helpful?  Adding button ID

johninaustin:
I'll setup VS Code & run tox / work on the tests.
Thanks for the review

johninaustin:
I've made the requested changes.  Can you please review?

mdonoughe (approved):
LGTM

bdraco:
Awesome! As soon as this gets released we can update Home Assistant to support picos without a pro bridge 👍

bdraco:
Tested this with https://github.com/home-assistant/core/pull/61032 and it works great

@mdonoughe Would you mind pushing a new release ?

Thanks!

mdonoughe:
I was already making the commit. :) https://pypi.org/project/pylutron-caseta/0.12.0/

swails:
Oh super cool.  I wonder when they added this feature to the bridge.

PR #87, opened 2022-02-01
Ensure bridge can still connect if there are no buttons

bdraco:
Fixes https://github.com/home-assistant/core/issues/65366


bdraco:
@mdonoughe Any chance we can get a new release with this today?  Sorry for the tight timeline, but it would be great to have this fixed before Home Assistant 2022.2 is released tomorrow

mdonoughe:
https://pypi.org/project/pylutron-caseta/0.13.1/

bdraco:
Thanks for the quick turn & adding the test 👍

PR #89, opened 2022-02-06
Add support for SerenaTiltOnlyWoodBlind

### LINKED ISSUE #88: Support for SerenaTiltOnlyWoodBlind
Please add support for SerenaTiltOnlyWoodBlind

swails:
Fixes #88 

I don't own one of these products, so I can't tell whether or not this is enough to fully support controlling the tilt angle.  Specifically I'm not sure if the LEAP messages are implemented or, if not, what those messages should be.


rlshaw23:
I have 4 of these. I'm no programmer but I will would be able to test this.

mdonoughe:
If we just add SerenaTiltOnlyWoodBlind to the same list as TriathlonRollerShade, how does the consuming code know that SerenaTiltOnlyWoodBlinds only support tilting and TriathlonRollerShades only support raising and lowering? I suspect with this code the blinds will appear in something like Home Assistant, but the default interaction will probably be to control the position that these blinds don't have.

It doesn't look like pylutron_caseta contains any code for controlling tilt and I expect it will be different from raising/lowering or else they wouldn't be able to sell any shades that have both raise/lower and tilt.

johninaustin:
I don't have shades to do any testing, but it seems there are different command types to work with different zones, and some include Tilt. 

    Raise,
    Lower,
    Stop,
    RaiseTilt,
    LowerTilt,
    StopTilt,
    GoToSwitchedLevel,
    GoToDimmedLevel,
    GoToShadeLevel,
    GoToShadeLevelWithTilt,
    GoToSpectrumTuningLevel,
    GoToReceptacleLevel,
    GoToCCOLevel,
    GoToFanSpeed,
    Activate,
    GoToGroupLightingLevel,
    GoToScene

If you do GoToShadeLevelWithTilt, it will take two parameters in the body - Tilt & Level.

When you query the zone, you should get Tilt the same way you get Level and FanSpeed. I expect setting works the same as well.

mdonoughe:
I think we're missing a GoToTilt command, but more than that this library does not have support for tilting shades at all. The tilt level is being reported as a Tilt property that needs to be read.

mdonoughe:
Using the leap command from #97, can you figure out how to read and control the tilt?

mdonoughe:
Can somebody install the latest development version with the cli extra (see [The leap tool](https://github.com/gurumitts/pylutron-caseta#the-leap-tool)) and try this command:

```
leap -X CreateRequest BRIDGE_IP/device/DEVICE_ID/commandprocessor -d '{"Command":{"CommandType":"GoToTilt","TiltParameters":{"Tilt":50}}}'
```

Maybe it will set the tilt to 50. My blinds are ZWave so I can't test this myself.

rlshaw23:
I’m running home assistant os. I tried to install the cli via putty. No go, pip not recognized.  I’m assuming that I won’t be able to test this command? I do have the Serena tilt only blinds.

mdonoughe:
I'm not sure how you would do it on Home Assistant OS. I've only ever used Home Assistant Core. If pip isn't found, you may need to use pip3 instead. It may be easier to run the cli from another computer.

It can be used from Windows but if all you have is Windows it may be easier using Python in WSL than Windows Python. I had problems with quotes in command line arguments. I don't know if it was PowerShell's native command quoting bugs or Python or both.

rlshaw23:
When I installed it this is the error message I get

pip3 install pylutron_caseta[cli]
Collecting pylutron_caseta[cli]
  Downloading pylutron_caseta-0.13.1-py3-none-any.whl (17 kB)
  WARNING: pylutron-caseta 0.13.1 does not provide the extra 'cli'

Any ideas what could be wrong. The leap tool will not work

mdonoughe:
The CLI hasn't been released yet. For now, you need to download the code from GitHub and do `pip install .[cli]` from inside the pylutron-caseta folder.

rlshaw23:
I cloned the directory and ran pip install .[cli] and this is what I get. see attached. Any thoughts?
error capture

mdonoughe:
I think you need to update pip.

DougieJynx:
> 
@mdonoughe just as an aside - I don't believe they sell any blinds that do both raise/lower and tilt, it's either/or for all of their products, so I wonder if they don't use the same functions for tilt as they do raise/lower.  I know the Lutron app allows you to set the boundaries for open and closed manually within the app, so maybe based on that they can send the same commands to the tilt blinds that they would send to the roller blinds?  As an additional aside I had the tilt blinds in SmartThings and they seem to integrate well there (although even if you integrate ST into HA it still won't bring over the blinds).

DougieJynx:
> -d '{"Command":{"CommandType":"GoToTilt","TiltParameters":{"Tilt":50}}}'


I'm not sure if this is helpful... I got a request back for TiltParameters when poking around with Tilt...

$:~/.config/pylutron_caseta# device=$(leap "${ip}/zone/status/expanded?where=Zone.ControlType:\"Tilt\"" | jq -r '.ZoneExpandedStatuses[0].Zone.href')
root@gemini:~/.config/pylutron_caseta# echo $device
/zone/21
$:~/.config/pylutron_caseta# leap -X CreateRequest "${ip}${device}/commandprocessor" -d '{"Command":{"CommandType":"GoToLevel","Parameter":[{"Type":"Level","Value":10}]}}'
{"Message": "GoToLevel not supported for zone type: Tilt"}
$:~/.config/pylutron_caseta# leap -X CreateRequest "${ip}${device}/commandprocessor" -d '{"Command":{"CommandType":"Tilt","Parameter":[{"Type":"Level","Value":10}]}}'
{"Message": "This command is not supported: Tilt"}
$:~/.config/pylutron_caseta# leap -X CreateRequest "${ip}${device}/commandprocessor" -d '{"Command":{"CommandType":"GoToTilt","Parameter":[{"Type":"Level","Value":10}]}}'
{"Message": "The body is missing a required property : TiltParameters"}

DougieJynx:
IT WORKED!!!!

$:~/.config/pylutron_caseta# leap -X CreateRequest "${ip}${device}/commandprocessor" -d '{"Command":{"CommandType":"GoToTilt","TiltParameters":{"Tilt":50}}}'
{"ZoneStatus": {"href": "/zone/21/status", "Tilt": 50, "Zone": {"href": "/zone/21"}, "StatusAccuracy": "Good", "Availability": "Available"}}
$:~/.config/pylutron_caseta# leap -X CreateRequest "${ip}${device}/commandprocessor" -d '{"Command":{"CommandType":"GoToTilt","TiltParameters":{"Tilt":25}}}'
{"ZoneStatus": {"href": "/zone/21/status", "Tilt": 25, "Zone": {"href": "/zone/21"}, "StatusAccuracy": "Good", "Availability": "Available"}}

mdonoughe:
Based on that, I added support for getting and setting the tilt, similar to how the existing code has been working. However, the way it works for projects like Home Assistant is not great. Basically, they know a cover supports tilt if `device['tilt'] is not None` and they know a cover supports raise/lower if `device['current_state'] != -1`.

mdonoughe:
I'm not happy with the current state of it but I released it so it can be used. On the consuming side, eg in Home Assistant, I guess it's possible to differentiate shades that tilt from shades that raise/lower by checking if the device has a tilt state, or by having a list of known devices that are tilting shades.

The bridge service has a concept of control types and that's how we should be detecting what operations are applicable for different devices. However, pylutron_caseta does not currently retrieve that property, and the property is associated with zones, while pylutron_caseta is currently device oriented. I think ultimately we need to be using that, but I haven't had time to set it up properly and I don't want to keep holding this up.

DougieJynx:
> If we just add SerenaTiltOnlyWoodBlind to the same list as TriathlonRollerShade, how does the consuming code know that SerenaTiltOnlyWoodBlinds only support tilting and TriathlonRollerShades only support raising and lowering? I suspect with this code the blinds will appear in something like Home Assistant, but the default interaction will probably be to control the position that these blinds don't have.
> 
> It doesn't look like pylutron_caseta contains any code for controlling tilt and I expect it will be different from raising/lowering or else they wouldn't be able to sell any shades that have both raise/lower and tilt.

They DON'T sell any shades that have both raise/lower and tilt... At least not currently.

Issue #93, opened 2022-02-19
async_pair: [Errno13] Permission Denied


michaelgaultjr:
Problem
---
Calling `await async_pair(SERVER_ADDRESS)` always returns a permission error regardless of the server address validity. However using [`get_lutron_cert.py`](https://github.com/gurumitts/pylutron-caseta/blob/dev/get_lutron_cert.py) with the same address works perfectly fine.
```
Traceback (most recent call last):
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.752.0_x64__qbz5n2kfra8p0\lib\concurrent\futures\_base.py", line 446, in result
    return self.__get_result()
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.752.0_x64__qbz5n2kfra8p0\lib\concurrent\futures\_base.py", line 391, in __get_result
    raise self._exception
  File "<console>", line 1, in <module>
  File "C:\Users\TGN-PC\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\LocalCache\local-packages\Python310\site-packages\pylutron_caseta\pairing.py", line 63, in async_pair
    csr, key_bytes_pem, ssl_context = await loop.run_in_executor(
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.752.0_x64__qbz5n2kfra8p0\lib\concurrent\futures\thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "C:\Users\TGN-PC\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\LocalCache\local-packages\Python310\site-packages\pylutron_caseta\pairing.py", line 205, in _generate_csr_with_ssl_context
    ssl_context.load_cert_chain(lap_cert_temp_file.name, lap_key_temp_file.name)
PermissionError: [Errno 13] Permission denied
```

Steps to reproduce
---
1. Open python REPL with asyncio module
```sh
$ python -m asyncio
```

2. Import `async_pair`
```python
>>> from pylutron_caseta.pairing import async_pair
```

3. Call `async_pair` with a valid or invalid server address
```python
>>> await async_pair("192.168.1.255")
```

Information
---
OS: `Windows 10 Pro`
OS Version: `21H1`
Python: `3.10.2`

[...]

PR #94, opened 2022-02-20
Pairing: Manually delete temporary files on windows

michaelgaultjr:
Changes
---
* Updated `pairing.py` to check if the current OS is Windows to manually delete the temporary files

Reason for Changes
---
Fixes the `PermissionError: [Errno 13] Permission denied` error that is being thrown when `async_pair` is called on Windows.

Related Issues
---
https://github.com/gurumitts/pylutron-caseta/issues/93

### DIFF
--- a/pylutron_caseta/pairing.py
+++ b/pylutron_caseta/pairing.py
@@ -236,4 +236,3 @@ def _generate_signed_ssl_context(key_bytes_pem, cert_pem, ca_pem):
                 os.remove(cert_temp_file.name)
 
             return signed_ssl_context
-            
--- a/pylutron_caseta/pairing.py
+++ b/pylutron_caseta/pairing.py
@@ -202,12 +202,12 @@ def _generate_csr_with_ssl_context():
 
                 ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
                 ssl_context.load_verify_locations(cadata=LAP_CA_PEM)
-                lap_key_temp_file.close()
                 ssl_context.load_cert_chain(lap_cert_temp_file.name, lap_key_temp_file.name)
                 ssl_context.verify_mode = ssl.CERT_REQUIRED
 
                 return csr, key_bytes_pem, ssl_context
             finally:
+                lap_key_temp_file.close()
                 lap_cert_temp_file.close()
                 os.remove(lap_key_temp_file.name)
                 os.remove(lap_cert_temp_file.name)
--- a/pylutron_caseta/pairing.py
+++ b/pylutron_caseta/pairing.py
@@ -6,6 +6,7 @@ import logging
 import socket
 import ssl
 import tempfile
+import os
 
 from cryptography import x509
 from cryptography.hazmat.backends import default_backend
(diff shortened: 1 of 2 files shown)


pylutron_caseta/pairing.py
```diff
@@ -187,8 +188,8 @@ async def _async_verify_certificate(server_addr, signed_ssl_context):
 
 
 def _generate_csr_with_ssl_context():
-    with tempfile.NamedTemporaryFile() as lap_cert_temp_file:
-        with tempfile.NamedTemporaryFile() as lap_key_temp_file:
+    with tempfile.NamedTemporaryFile(delete=not IS_WINDOWS) as lap_cert_temp_file:
```
mdonoughe, line 191: The Windows code works for all operating systems. It's easier to only write one version of the code.

pylutron_caseta/pairing.py
```diff
@@ -204,16 +205,22 @@ def _generate_csr_with_ssl_context():
             ssl_context.load_verify_locations(cadata=LAP_CA_PEM)
             ssl_context.load_cert_chain(lap_cert_temp_file.name, lap_key_temp_file.name)
             ssl_context.verify_mode = ssl.CERT_REQUIRED
+            
+            if (IS_WINDOWS):
+                lap_key_temp_file.close()
```
mdonoughe, line 210: I don't think this will solve your problem. The file needs to be closed before the call to `ssl_context.load_cert_chain`. That's really the problem with NamedTemporaryFile. Closing the file so it can be opened for reading again causes it to be deleted.

pylutron_caseta/pairing.py
```diff
@@ -204,16 +205,22 @@ def _generate_csr_with_ssl_context():
...
+                lap_cert_temp_file.close()
+                os.remove(lap_key_temp_file.name)
```
mdonoughe, line 212: The temp file should probably be deleted in a finally block.

mdonoughe (changes requested):

pylutron_caseta/pairing.py
```diff
@@ -6,6 +6,7 @@
 import socket
 import ssl
 import tempfile
+import os
```
mdonoughe, line 9: You don't need this import anymore.

pylutron_caseta/pairing.py
michaelgaultjr: @mdonoughe  it seems to be working fine being open until that point, though I don't have any other ways of testing it at the moment so I don't know if it's just me

mdonoughe: I'd expect this code to work the same as the original code unless setting delete has some interaction with file share modes.

pylutron_caseta/pairing.py
mdonoughe: Oh actually you do need this because you're using `os.remove`.

pylutron_caseta/pairing.py
mdonoughe: You are right. I guess there is an interaction with file sharing.

michaelgaultjr: @mdonoughe just did a little more testing, and there doesn't appear to be a difference calling it before or after ` ssl_context.load_cert_chain` I'll leave it before just incase it's just me.

mdonoughe: Actually you probably want to leave it after if that works. Otherwise the error handling becomes more complicated because the file always needs to be closed before deletion. In any case, the handling for the two files should be consistent.

PR #97, opened 2022-04-03
create cli tools

mdonoughe:
This PR replaces get_lutron_cert.py with a lap-pair tool and adds two additional tools: leap-scan and leap. This should make it easier to probe around the LEAP server and add new capabilities to pylutron_caseta.


pylutron_caseta/cli.py
```diff
@@ -0,0 +1,342 @@
...
+
+    By default, the certificate files will be placed in
+    $XDG_CONFIG_HOME/pylutron_caseta, named after the address of the bridge device. The
+    leap tool will look for certificates in the same location.
+    """
+
+    def _ready():
+        click.echo(
+            "Press the small black button on the back of the bridge to complete "
+            + "pairing."
```
swails, line 129:
```suggestion
            "pairing."
```

The + isn't needed (and without it python builds a single string). With the + python builds 2 strings, then adds them building a third).

pylutron_caseta/cli.py
```diff
@@ -0,0 +1,342 @@
...
+
+@click.command()
+@_cli_main
+@click.option(
+    "-i",
+    "--interface",
+    multiple=True,
+    help=(
+        "Limit scanned network interfaces. "
+        + "This option may be specified multiple times."
```
swails, line 147:
```suggestion
        "This option may be specified multiple times."
```

swails (approved):
Looks nice! Definitely useful

PR #99, opened 2022-05-19
Link Occupancy sensor device to Occupancy group.

bpaauwe:
Without a link between the two, there is no way to know which
device reported the occupancy status.

The device information contrains a list of OccupancySensors as
does the Occupancy group information.  So let's add the lists
into our public data structures so the link can be established.

For the device, we use the same method as button_groups to create
a new field called occupancy_sensors.

For the occupancy group we create an array from the
AssociatedSensors and add it as a new field called sensors.


swails (approved):

swails:
Unit tests need to be updated here

bpaauwe:
I think I updated the unit tests correctly.

PR #107, opened 2022-08-17
Add occupancy sensor support for RA3 processors

danaues:
This PR adds support for occupancy sensors for RA3 processors

The RA3 processor does not allow us to query the /occupancygroup URL like we can with Caseta.

I can query all the devices and filter by DeviceType "RPSOccupancySensor" to discover the sensors, and then extract the associated areas.  We can also query /area/status , but the processor returns areas that do not have sensors, so we end up with occupancy groups that will never have a status other than Unknown.  The DeviceType method ensures we are only exposing areas that have a sensor.

Subscribing to /area/status allows us to retrieve occupancy updates from the processor.

Tested on RA3 Processor, and Caseta Pro bridge.
I only have 1 occupancy sensor, but I've simulated multiple sensors, including 2 sharing an area in the tests.


pylutron_caseta/smartbridge.py
```diff
@@ -565,6 +566,37 @@ def _handle_occupancy_group_status(self, response: Response):
             if occgroup_id in self._occupancy_subscribers:
                 self._occupancy_subscribers[occgroup_id]()
 
+    def _handle_ra3_occupancy_group_status(self, response: Response):
+        _LOG.debug("Handling ra3 occupancy status: %s", response)
+
+        if response.Body is None:
+            return
+
+        statuses = response.Body.get("AreaStatuses", {})
```
mdonoughe, line 575:
```suggestion
        statuses = response.Body.get("AreaStatuses", [])
```
This looks like it's supposed to be a list, not a dictionary.

pylutron_caseta/smartbridge.py
```diff
@@ -989,6 +1023,70 @@ def _process_occupancy_group(self, occgroup):
             name=f"{self.areas[occgroup_area_id]['name']} Occupancy",
         )
 
+    async def _load_ra3_occupancy_groups(self):
+        """Load the devices from the bridge and filter for occupancy sensors."""
+        _LOG.debug("Finding occupancy sensors from bridge")
+        occdevice_json = await self._request(
+            "ReadRequest", "/device?where=IsThisDevice:false"
```
mdonoughe, line 1030: Is there a way to query just the occupancy sensors? If there is only one kind of occupancy sensor we could filter by type. The test data doesn't give many hints.

pylutron_caseta/smartbridge.py
```diff
@@ -989,6 +1023,70 @@ def _process_occupancy_group(self, occgroup):
...
+                "Occupancy area %s already registered, append sensor %s",
+                occgroup_area_id,
+                occdevice_id,
+            )
+            self.occupancy_groups[occgroup_area_id]["sensors"].append(occdevice_id)
+            return
+
+        occsensor_ids.append(occdevice_id)
+
+        self.occupancy_groups.setdefault(
```
mdonoughe, line 1066: This function seems like maybe it was partially refactored.

At the top of the function, a variable `occsensor_ids` is initialized to `[]`. This variable is only read once and only after it is updated to `[occdevice_id]`. It could be replaced with just `[occdevice_id]`.

This code here uses `self.occupancy_groups.setdefault(occgroup_area_id, …)`, but the code just before it does an early return if `occgroup_area_id in self.occupancy_groups`. So this code could be simplified to `self.occupancy_groups[occgroup_area_id] = { occupancy_group_id: occgroup_area_id, status: OCCUPANCY_GROUP_UNKNOWN, sensors: occsensor_ids, name: f"{self.areas[occgroup_area_id]['name']} Occupancy" }`.

Alternatively, the return value of `self.occupancy_groups.setdefault(occgroup_area_id, { …, sensors: [], })` could be stored and then `whatever.sensors.append(occ_device_id)` would add the sensor to the list, but then the logging would be different.

mdonoughe (changes requested):
Mostly good

pylutron_caseta/smartbridge.py
danaues: Unfortunately, I have not been able to find a way to query just the occupancy sensors.  This is the only way I've found to discover them at all.

pylutron_caseta/smartbridge.py
danaues: Yes, I agree.  I've removed the unused variable occsensor_ids, and taken your advice with returning the value of self.occupancy_groups.setdefault(..) and appending the sensor.  The early check and return has been removed.

danaues:
Thx!

PR #109, opened 2022-08-30
RA3/QSX - Add support for RPSCeilingMountedOccupancySensor

danaues:
This PR adds support for a new occupancy sensor device type that has been discovered.

A user that has been helping with testing has a sensor with DeviceType "RPSCeilingMountedOccupancySensor"

Adding this sensor type to RA3_OCCUPANCY_SENSOR_DEVICE_TYPES enables this occupancy sensor.


bdraco:
FYI, today is the last day we can take a PR to bump the lib in HA if we want to get RA3 support in beta

PR #118, opened 2022-09-27
Update RA3/HWQSX naming to be consistent with Caseta naming

danaues:
This update changes the naming of RA3 devices and keypads to match the Caseta naming convention.

1. "_" character only appears between area and the rest of the name, so the integration can parse out the area from the names, but the names will not contain underscores, just spaces.  Not implemented for Processor or Occupancy Group, as these are handled differently by the integration for now.  This is consistent with Caseta.

2. Keypads and Picos were adding the full DeviceType to the name, where in Caseta we just add the word Pico.  Now all Keypads will say just "Keypad" and Picos will say "Pico", instead of the full DeviceType (SunnataKeypad).


danaues:
Closing, some errant commits included by accident.

Issue #122, opened 2022-10-06
support for ra3 phantom keypads?


JonGilmore:
Thanks for all of the hard work on this! Is it possible to get this to support phantom keypads? 

<img width="736" alt="image" src="https://user-images.githubusercontent.com/7232986/194191762-9cad7d40-e23c-4304-a022-ecc8f180e84c.png">

The big benefit of using phantom keypads (which I previously used in ra2 with pylutron) is that you can create all kinds of scenes to trigger without using actual buttons on your physical keypads. Here's an example:

<img width="585" alt="image" src="https://user-images.githubusercontent.com/7232986/194192127-fd427152-6b93-4166-9e68-7ebd50a13bca.png">

danaues:
@JonGilmore I was just thinking about this the other day.

Shouldn't be a problem to add support for this and the HomeOwner Keypad.  The key is that neither of these keypads supply a serial number, so we just have to work around that.

I'll address it this week.

danaues:
Need to add support in 2 places.

This adds support to pylutron-caseta
https://github.com/gurumitts/pylutron-caseta/pull/123

Second is the lutron_caseta HA integration.  I'll PR the Blank serial handling for this as soon as I'm able.

I'll let you know when you can help test it.

swamplynx:
@danaues will this also allow HA to "push" buttons on the keypads, or is there more work to be done there? I know it is / was on your roadmap. Definitely a critical feature for me so that HA can execute Lutron scenes.

danaues:
@swamplynx 
already coded and working, just have to PR into HA one feature at a time.  It's next in line after adding this keypad support.

danaues:
@JonGilmore how do press/trigger these buttons in your RA3 system?  The PhantomKeypad doesn't show up in the Lutron App on my phone.  Makes it difficult to test the button presses.

danaues:
https://github.com/gurumitts/pylutron-caseta/pull/123 has been merged adding support to pylutron-caseta

Home Assistant PR79958 has been created to add support to HA
https://github.com/home-assistant/core/pull/79958

danaues:
https://github.com/home-assistant/core/pull/79958 has been merged as well

@JonGilmore would you please close this Issue, as implementation is now complete.

Thanks.

danaues:
@swamplynx 

Home assistant PR 79963 for button presses
https://github.com/home-assistant/core/pull/79963

JonGilmore:
[...]

PR #123, opened 2022-10-06
Add support for PhantomKeypad

danaues:
Adding support for the PhantomKeypad DeviceType.

The PhantomKeypad is the RA2 version of the RA3 HomeOwner Keypad. 
Both of these are virtual Keypads.

DeviceType has been confirmed on my own system.


danaues:
Resolves this issue:
https://github.com/gurumitts/pylutron-caseta/issues/122

PR #124, opened 2022-10-07
look for _lutron._tcp.local

mdonoughe:
Apparently the _leap service is not used with the newer hardware. Home Assistant has its own version of this and should make the same change.


danaues:
ok, agreed.  I'll take care of it in the lutron_caseta integration.

danaues:
PR to HomeAssistant for this change
https://github.com/home-assistant/core/pull/80852

PR #129, opened 2022-10-19
Fix area tree discovery

danaues:
Currently, the area discovery ignores any area that has the field IsLeaf: False

Unfortunately, this leads to a broken area tree for RA3/HWQSX if the area's don't contain any control stations or devices.

In the example below, both "tree 1" and "tree 2" areas are reported as having "None" parent, because the isLeaf field is False as a consequence of not containing any devices/control stations.
"tree 3" correctly reports it's parent as "tree 2"

image

This update ignores the IsLeaf field, and discovers all areas defined on the bridge/processor, maintaining the correct parent/child relationship.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Now discovers the complete area list with working parent_area relationship.
+
 ## [0.17.0] - 2022-10-18
 
 ### Added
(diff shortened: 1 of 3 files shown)


mdonoughe:
Can you explain this change in the changelog?

danaues:
Yes, no problem

danaues:
Thanks!

danaues:
If you are ok release this as 0.71.1, I'll PR the bump to HA.

mdonoughe:
https://pypi.org/project/pylutron-caseta/0.17.1/

danaues:
Thanks!

PR #130, opened 2022-10-27
Add gang count and position to ra3/hwqsx control stations

danaues:
Adding 2 fields into the devices dict for RA3/HWQSX keypads/picos:
gang_position (which position the keypad resides in a multi-gang station)
gang_count (how many ganged devices are in the control station)

These fields will allow for better device/entity naming.

Currently each keypad has "Position 1" in it's name, even if it's the only keypad in that location. Even a single pico on a pedestal gets the "Position 1" added to it's name.

With these fields, we can detect if it's in a single gang, and ignore the position name.

ie "Rec Room Coffee Table Position 1 Pico" becomes "Rec Room Coffee Table Pico"

The gang_position field will allow us to convert the position to an ordinal (1st, 2nd etc) without having to parse the number out of the device_name field.

### DIFF
--- a/pylutron_caseta/smartbridge.py
+++ b/pylutron_caseta/smartbridge.py
@@ -851,14 +851,14 @@ class Smartbridge:
         for station in station_json:
             station_name = station["Name"]
             ganged_devices_json = station["AssociatedGangedDevices"]
-
+        
             for device_json in ganged_devices_json:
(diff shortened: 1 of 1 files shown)


iCSpotRun:
@danaues I would love to see this PR get merged. Dealing with a setup where many multi-gangs

:)

PR #131, opened 2022-10-31
Update __init__.py

B1223GS87:
Add newly introduced Sunnata Hybrid Keypad.


danaues:
Do you have one of these to verify that is the correct DeviceType?
I'm not able to add this into my Lutron Designer yet to verify.

mdonoughe:
This seems like it's right.

PR #138, opened 2023-03-07
Switch to using asyncio.timeout instead of asyncio.wait_for

bdraco:
`asyncio.wait_for` creates a task whereas `asyncio.timeout` avoids doing this which decreases latency.

Fallback to using `async_timeout` when the python version is too old (<3.11)

`asyncio.timeout` will become the underlying implementation for `async.wait_for` in cpython 3.12 https://github.com/python/cpython/pull/98518


pylutron_caseta/pairing.py
```diff
@@ -220,9 +218,9 @@ async def _async_verify_certificate(server_addr, signed_ssl_context):
             return leap_response
 
 
-def _generate_csr_with_ssl_context() -> Tuple[
-    x509.CertificateSigningRequest, bytes, ssl.SSLContext
-]:
+def _generate_csr_with_ssl_context() -> (
```
bdraco, line 221: black seems to want this even though I think it looks worse 🤷‍♂️

bdraco:
thanks

PR #145, opened 2023-08-21
Add device type for Wall Mounted Motion Sensor in RA3 Firmware v23.4

swamplynx:
The RA3 firmware that ships with the latest v23.4 Designer software changes the device type of wall mounted occupancy sensors in LEAP to "RPSWallMountedOccupancySensor," which was previously not a part of this library. This changes adds that device type. The new firmware also appears to change the device type of Phantom Keypads, however my debug logs only show device type as "Unknown," so someone with more architecture knowledge may need to step in and help with that change.


swamplynx:
Thanks for merging @mdonoughe. Can we publish to PyPi as v0.18.2?

mdonoughe:
https://pypi.org/project/pylutron-caseta/0.18.2/

Issue #149, opened 2023-09-24
Add Support for Lumaris Tape Light


RBaragona:
Add support for https://luxury.lutron.com/us/en/lighting/lumaris-tunable-white-tape-light

```
"8157": {
          "device_id": "8157",
          "current_state": 0,
          "fan_speed": null,
          "zone": "8157",
          "name": "Kitchen_Island Toe Kick",
          "button_groups": null,
          "type": "WhiteTune",
          "model": null,
          "serial": null,
          "area": "587",
          "device_name": "Island Toe Kick",
          "tilt": null
```

I attempted to add this myself (https://github.com/RBaragona/pylutron-caseta/tree/Support-Lumaris-Tape-Light), following https://github.com/gurumitts/pylutron-caseta/issues/115#issuecomment-1250413666, but my version continued to downgrade to 0.18.2.  Any advice for adding a development version would be appreciated.

mdonoughe:
I don't think those instructions work. You should install the library into config/deps. Remember to uninstall it later or you'll still be using the config/deps version even when the version included in Home Assistant is newer.

RBaragona:
Thanks for the quick repsonse, @mdonoughe .  Unfortunately, I'm not familiar with that process and couldn't quickly find any documentation.  Can you point to any instructions or steps I can take to install the library to config/deps?

mdonoughe:
I tried writing documentation for it: https://github.com/mdonoughe/developers.home-assistant/blob/lib-test/docs/api_lib_testing.md

RBaragona:
Thanks for this write up, it’s very helpful!  I think the issue I’m having is with wheel compilation as you’ve stated. I can compile/install if I don’t specify the does folder. Once I do specify deps folder, compiling fails for cffi. I’ll need to do some more research in that direction.

RBaragona:
Finally had enough time to dig into getting this setup, per @mdonoughe 's instructions, I wasn't able to cross compile in to production setup.  I needed the time to learn how to setup a disposable Home Assistant installation by following the [development environment instructions](https://developers.home-assistant.io/docs/development_environment).

mdonoughe:
Fixed in #154

PR #154, opened 2023-12-04
Add Lumaris Tape Light and Ketra color support

eclair4151:
Here's my initial attempt at adding ketra color support. Includes unit tests, and supports zone status for retrieving the color of the lamp. Let me know if you think something should be changed. Tested and works great on my Ketra bulbs, but can use further confirmation. 
See issue #153 

Corresponding Home Assistant feature branch:
https://github.com/eclair4151/home-assistant-core/tree/ketra_lumaris_color_support

Once you checkout the HA branch, youll need to manually install the new version of pylutron-caseta as well.
Open a terminal in your Home Assistant Core, or the terminal tab in vscode docker, and run

pip uninstall pylutron-caseta    
pip install git+https://github.com/eclair4151/pylutron-caseta.git@dev 

Video of this working in Home Assistant
https://photos.app.goo.gl/VAziwYKXhbo8fo4z7

PR #155, opened 2023-12-13
Add Lumaris Tape Light Support

RBaragona:
Adding support for dimming Lumaris Tape Light, for issue #149

### DIFF
--- a/pylutron_caseta/smartbridge.py
+++ b/pylutron_caseta/smartbridge.py
@@ -7,6 +7,7 @@ import math
 import socket
 import ssl
 from typing import Callable, Dict, List, Optional, Tuple, Union
+from .color_value import ColorValue
 
 try:
     from asyncio import get_running_loop as get_loop
@@ -281,14 +282,15 @@ class Smartbridge:
         return (response, tag)
 
     async def set_value(
-        self, device_id: str, value: int, fade_time: Optional[timedelta] = None
+        self, device_id: str, value: Optional[int] = None, fade_time: Optional[timedelta] = None, color_value: Optional[ColorValue] = None
     ):
         """
         Will set the value for a device with the given ID.
 
         :param device_id: device id to set the value on
-        :param value: integer value from 0 to 100 to set
+        :param value: integer value from 0 to 100 to set (Optional if just setting color)
         :param fade_time: duration for the light to fade from its current value to the
+        :param color_value: color value to set the device to (only currently valid for Ketra devices)
         new value (only valid for lights)
         """
         device = self.devices[device_id]
@@ -310,7 +312,11 @@ class Smartbridge:
 
         # Handle Ketra lamps
         if device.get("type") == "SpectrumTune":
-            params = {"Level": value}  # type: Dict[str, Union[str, int]]
+            params = {}  # type: Dict[str, Union[str, int]]
+            if value is not None:
+                params["Level"] = value
+            if color_value is not None:
+                params.update(color_value.get_spectrum_tuning_level_parameters())
             if fade_time is not None:
                 params["FadeTime"] = _format_duration(fade_time)
             await self._request(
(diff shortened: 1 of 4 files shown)


mdonoughe:
I think there might be an issue with the approach taken by with WhiteTune and the existing SpectrumTune. Is it possible to control the brightness of the light anymore or just the color temperature? Maybe it should be more like #154 .

RBaragona:
This does control the brightness of the Lumaris lights in my HomeWorks QSX system. To gain white tuning a similar approach to #154 would be needed. I was waiting to add that feature until the Keyra capability was reviewed to follow the same pattern and any potentially associated changes.

pylutron_caseta/smartbridge.py
```diff
@@ -324,6 +324,23 @@ async def set_value(
                 },
             )
             return
+        
+        # Handle Lumaris Tape Light
+        if device.get("type") == "WhiteTune":
+            params = {"Level": value}  # type: Dict[str, Union[str, int]]
```
eclair4151, line 330: My PR #154 includes the code for white tuning color changes by setting the kelvin value. 
All that would be needed to support color tuning after my PR is merged would be to change line 330
`params = {"Level": value}
`
To update it to the same as my PR. 
```
params = {}  # type: Dict[str, Union[str, int]]
if value is not None:
    params["Level"] = value
if color_value is not None:
    params.update(color_value.get_spectrum_tuning_level_parameters())
```

RBaragona: @eclair4151  do you have the POC changes to home assistant available in your fork? I’d like to test them out and confirm Lumaris wouldn’t get color sent, only temperature. This is further complicated since there are two kelvin range models that I still need to determine home to differentiate and set kelvin limits.

eclair4151: Ah ok, I believe there is a leap command to pull the kelvin limits from the zone, which we would need to implement to correctly set the scale for each device. 

In terms of the HA fork, It's not quite ready yet, but after I saw your PR, I did start adding support in. 

In the HA plugin you define for each device which capabilities it supports. 
https://developers.home-assistant.io/docs/core/entity/light/

So for the WhiteTune zone, it will define
  _attr_color_mode = ColorMode. COLOR_TEMP
  _attr_supported_color_modes = {
      ColorMode.BRIGHTNESS,
      ColorMode.COLOR_TEMP,
      ColorMode.HS,
  }

eclair4151: @RBaragona I see the zone expanded status leap call returns an object which includes 
```
"ColorTuningProperties": {
            "WhiteTuningLevelRange": {
              "Max": 10000,
              "Min": 1400
            }
          },
```

only question is Lutron app artificially caps the kelvin limit at 7K on the front end, while my bulb lists as supporting 10k. Do we want to do the same on our end and cap it at 7k, or just ignore that, and directly pass in the limit provided over leap?

RBaragona: I'm not sure how you were able to get the expanded status, but I'd like to do that with the Lumaris devices.

With my Ketra devices, it does appear 10k was achievable.  LEAP protocol states it was set to 10k, but the Lutron app shows it in color tune mode, not white tuning.  Interestingly, the Lumaris devices also accepted the 10k value, but fed back their level as the max allowed (ie I set 10k, but the device is 3k max, so it went to 3k and reported 3).  I did recieve a LEAP error for Ketra devices and Lumaris beyond 10k.

eclair4151: I'll let you know when its ready to test. I was also thinking if it makes sense to combine our PR's into 1 since they are pretty dependent on each other.

RBaragona: Sounds like a good plan to me!

uberjay:
Hi, came here looking for Lumaris support with my RA3 system. I'd be happy to test whatever changes you've got -- I already have a HA dev-container env set up, just waiting to be pointed at a branch. 😁

eclair4151:
@uberjay  Great, ill let you know as soon as my HA branch is ready to test, we're just finalizing up some details on implementing the kelvin limits for each device.

eclair4151:
@uberjay @RBaragona i've gone ahead and merged your branch into mine and pushed, so should be good to close this one. Luckily the white tuning range was already being returned in the associatedzone call during zone loading, so no extra leap call is needed, I just added in the parsing. 

You should be able to check out my branch and check the warm cool limits on your device by running:

```
    await bridge.connect()
    device_id = "XYZ"
    print(bridge.get_device_by_id(ketra_device_id)["white_tuning_range"])
```

ill let you know as soon as the HA branch is ready.

mdonoughe:
Superseded by #154

Issue #158, opened 2024-01-30
devcontainer build: file permissions error on updateContent.sh makes the devcontainer build fail


dkulla01:
I ran into an issue trying to build the devcontainer for this project. I looked through the build logs to try to figure out what was going on, and I saw a file permissions error:

```
Running the updateContentCommand from devcontainer.json...

[3955 ms] Start: Run in container: /bin/sh -c ./.devcontainer/updateContent.sh
/bin/sh: 1: ./.devcontainer/updateContent.sh: Permission denied
[4087 ms] updateContentCommand failed with exit code 126. Skipping any further user-provided commands.
[4090 ms] Error: Command failed: /bin/sh -c ./.devcontainer/updateContent.sh
[4090 ms]     at fY (/Users/dan/.vscode/extensions/ms-vscode-remote.remote-containers-0.330.0/dist/spec-node/devContainersSpecCLI.js:235:130)
[4090 ms]     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
[4090 ms]     at async rl (/Users/dan/.vscode/extensions/ms-vscode-remote.remote-containers-0.330.0/dist/spec-node/devContainersSpecCLI.js:227:4393)
[4090 ms]     at async el (/Users/dan/.vscode/extensions/ms-vscode-remote.remote-containers-0.330.0/dist/spec-node/devContainersSpecCLI.js:227:3738)
[4090 ms]     at async il (/Users/dan/.vscode/extensions/ms-vscode-remote.remote-containers-0.330.0/dist/spec-node/devContainersSpecCLI.js:227:2814)
[4090 ms]     at async TeA (/Users/dan/.vscode/extensions/ms-vscode-remote.remote-containers-0.330.0/dist/spec-node/devContainersSpecCLI.js:614:30432)
[4090 ms]     at async PeA (/Users/dan/.vscode/extensions/ms-vscode-remote.remote-containers-0.330.0/dist/spec-node/devContainersSpecCLI.js:614:27664)
[4096 ms] Exit code 1
[...]

PR #159, opened 2024-01-30
invoke bash ./.devcontainer/updateContent.sh

dkulla01:
without "bash" at the start of this command, the devcontainer machinery invokes "sh -c ./.devcontainer/updateContent.sh". 

But this file is not executable, so we get a permissions denied error. Changing the command to

```
bash ./.devcontainer/updateContent.sh
```

tells the build machinery to invoke `bash` (which is executable) and use it to read and execute the specified script. Since `updateContent.sh` is readable, and since `bash` is executable, the build machinery is able to run the `updateContentCommand` script this way.

see https://github.com/gurumitts/pylutron-caseta/issues/158

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -17,5 +17,5 @@
 			]
 		}
 	},
-	"updateContentCommand": "./.devcontainer/updateContent.sh"
+	"updateContentCommand": "bash ./.devcontainer/updateContent.sh"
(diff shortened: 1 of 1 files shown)


mdonoughe:
I should have marked the file as executable, but maybe editing on Windows messed things up. Invoking it with bash should work even if somebody downloads the source in a zip file so it's probably better.

Issue #152, opened 2023-11-30
KeyError: No device associated with zone X


jonoberheide:
Hi there! Using lutron_caseta as part of Home Assistant. When integrating with a QSX bridge, I get the following error:

```
2023-11-30 13:07:33.288 ERROR (MainThread) [homeassistant.config_entries] Error setting up entry 192.168.30.20 for lutron_caseta
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/config_entries.py", line 402, in async_setup
    result = await component.async_setup_entry(hass, self)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/lutron_caseta/__init__.py", line 176, in async_setup_entry
    await bridge.connect()
  File "/usr/local/lib/python3.11/site-packages/pylutron_caseta/smartbridge.py", line 106, in connect
    await self._login_completed
  File "/usr/local/lib/python3.11/site-packages/pylutron_caseta/smartbridge.py", line 692, in _login
    await self._load_ra3_devices()
  File "/usr/local/lib/python3.11/site-packages/pylutron_caseta/smartbridge.py", line 801, in _load_ra3_devices
    await self._subscribe_to_multi_zone_status()
  File "/usr/local/lib/python3.11/site-packages/pylutron_caseta/smartbridge.py", line 1291, in _subscribe_to_multi_zone_status
    self._handle_multi_zone_status(response)
  File "/usr/local/lib/python3.11/site-packages/pylutron_caseta/smartbridge.py", line 603, in _handle_multi_zone_status
    self._handle_zone_status(zonestatus)
  File "/usr/local/lib/python3.11/site-packages/pylutron_caseta/smartbridge.py", line 552, in _handle_zone_status
    device = self.get_device_by_zone_id(zone)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pylutron_caseta/smartbridge.py", line 201, in get_device_by_zone_id
    raise KeyError(f"No device associated with zone {zone_id}")
KeyError: 'No device associated with zone 5435'
```

[...]

PR #161, opened 2024-02-08
add LEAP Paging/ContinuationToken support

jonoberheide:
LEAP supports a pagination mechanism when it hits limits on the number of objects returned in a particular API call. I observed this when retrieving more than 99 areas on a QSX system here: https://github.com/gurumitts/pylutron-caseta/issues/152

The response header will contain a Paging object which has a ContinuationToken that should be reflected back into a subsequent request to continue returning additional results.

This PR adds paging support to both the leap client, for manual paging invocation, and the smartbridge interface, by repeating requests until all pagination is complete and the results are merged together.

### DIFF
--- a/src/pylutron_caseta/cli.py
+++ b/src/pylutron_caseta/cli.py
@@ -321,7 +321,7 @@ async def leap(
 
     if verbose:
         # LeapProtocol discards the original JSON so reconstruct it here.
-        header = {
+        message = {
             "Header": {
                 "StatusCode": str(response.Header.StatusCode),
                 "Url": response.Header.Url,
@@ -331,9 +331,9 @@ async def leap(
             "Body": response.Body,
         }
         if response.Header.Paging:
-            header["Header"]["Paging"] = response.Header.Paging
+            message["Header"]["Paging"] = response.Header.Paging
 
-        output.write(json.dumps(header))
+        output.write(json.dumps(message))
     else:
         output.write(json.dumps(response.Body))
 
--- a/src/pylutron_caseta/cli.py
+++ b/src/pylutron_caseta/cli.py
@@ -261,6 +261,7 @@ async def _connect(
     help="The path to the client certificate key.",
 )
 @click.option("-d", "--data", help="The JSON data to send with the request.")
+@click.option("-p", "--paging", help="The JSON Paging dict to send with the request.")
 @click.option(
     "-f",
     "--fail",
@@ -286,6 +287,7 @@ async def leap(
     cert: str,
     key: str,
     data: Optional[str],
+    paging: Optional[str],
     fail: bool,
     output: TextIO,
     verbose: bool,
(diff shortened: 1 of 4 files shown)


src/pylutron_caseta/smartbridge.py
```diff
@@ -251,14 +251,32 @@ async def _request(
...
+            responses.append(response)
+
+            paging = response.Header.Paging
+            if not paging:
+                break
+
+        # merge the Body of multiple paged Responses together
+        merged = responses.pop(0)
+        if merged.Body:
+            merged_type = list(merged.Body.keys())[0]
```
mdonoughe, line 275: I don't know if this is quite right. Maybe we should merge every value which contains a list. The problem with taking the first one is that the order of JSON properties usually has no meaning.

src/pylutron_caseta/cli.py
```diff
@@ -316,19 +321,19 @@ async def leap(
...
-                        "StatusCode": str(response.Header.StatusCode),
-                        "Url": response.Header.Url,
-                        "MessageBodyType": response.Header.MessageBodyType,
-                    },
-                    "CommuniqueType": response.CommuniqueType,
-                    "Body": response.Body,
-                }
-            )
-        )
+        header = {
```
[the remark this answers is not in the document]
jonoberheide: Not quite the Header, definitely not the Body. 😄 `leap`? `preamble`?

src/pylutron_caseta/smartbridge.py
jonoberheide: Yeah, hard to know without authoritative protocol documentation. I'm not sure if there are responses that contain multiple entity types within the Body, but merging all the lists with the same top-level entity type seems reasonable?

src/pylutron_caseta/cli.py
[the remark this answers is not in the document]
mdonoughe: This is the reconstructed JSON response. `response` is already taken by the response model.

jonoberheide:
Any further thoughts on the above comments or the PR, @mdonoughe @gurumitts?

src/pylutron_caseta/cli.py
```diff
@@ -316,19 +321,19 @@ async def leap(
...
+                "Url": response.Header.Url,
+                "MessageBodyType": response.Header.MessageBodyType,
+                },
+            "CommuniqueType": response.CommuniqueType,
+            "Body": response.Body,
+        }
+        if response.Header.Paging:
+            header["Header"]["Paging"] = response.Header.Paging
+
+        output.write(json.dumps(header))
```
mdonoughe, line 336:
```suggestion
        message = {
            "Header": {
                "StatusCode": str(response.Header.StatusCode),
                "Url": response.Header.Url,
                "MessageBodyType": response.Header.MessageBodyType,
                },
            "CommuniqueType": response.CommuniqueType,
            "Body": response.Body,
        }
        if response.Header.Paging:
            message["Header"]["Paging"] = response.Header.Paging

        output.write(json.dumps(message))
```

PR #168, opened 2024-06-24
Fix license classifier

joostlek:
Hey 👋🏻,

I am currently looking into licensing at Home Assistant, and I found that we could not detect the license of this library properly. According to pyproject.toml documentation about the `license` field:
> If you are using a standard, well-known license, it is not necessary to use this field. Instead, you should use one of the [classifiers](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#classifiers) starting with License ::. (As a general rule, it is a good idea to use a standard, well-known license, both to avoid confusion and because some organizations avoid software whose license is unapproved.)

If you could do a release after this PR, that would be awesome :)

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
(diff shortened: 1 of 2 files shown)


joostlek:
Also, please consider adding a LICENSE file to the root of the project so its easier to see that its OSI approved

mdonoughe (changes requested):
This PR changes the license of the project.

This project is licensed under the Apache-2.0 license, and that is [correctly specified](https://hatch.pypa.io/1.9/config/metadata/#license) in the pyproject.toml file and package metadata. `License :: OSI Approved :: Apache Software License` is one of the old classifiers on the [PEP 639 blacklist](https://peps.python.org/pep-0639/appendix-mapping-classifiers/#spec-mapping-classifiers-identifiers).

I guess if they're both specified it's clear what the real license is, and since the classifier name contains "OSI Approved" it's one of the OSI approved versions of Apache so that's probably good enough, but a Home Assistant license check script should understand PEP 639 license metadata because at least Poetry and Hatch recommend using standard SPDX licenses (PEP 639) instead of Python classifiers. It looks like Poetry will add this ambiguous classifier, but Hatch doesn't add license classifiers.

mdonoughe (approved):

joostlek:
Thanks for wrapping it up. I didn't completely understand what direction you wanted to go in. But much appreciated 😊

PR #175, opened 2024-11-14
use orjson instead of json

jonoberheide:
Use orjson instead of json like home assistant does. pylutron-caseta spends a lot of time munging json and orjson is fast at that. The main compatibility difference between json and orjson is that dumps() returns bytes.

### DIFF
--- a/src/pylutron_caseta/leap.py
+++ b/src/pylutron_caseta/leap.py
@@ -85,7 +85,7 @@ class LeapProtocol:
             if received == b"":
                 break
 
-            resp_json = orjson.loads(received.decode("UTF-8"))
+            resp_json = orjson.loads(received)
 
             if isinstance(resp_json, dict):
                 tag = resp_json.get("Header", {}).pop("ClientTag", None)
--- a/src/pylutron_caseta/leap.py
+++ b/src/pylutron_caseta/leap.py
@@ -1,11 +1,12 @@
 """LEAP protocol layer."""
 
 import asyncio
-import orjson as json
 import logging
 import re
 import uuid
 from typing import Callable, Dict, List, Optional, Tuple
+
+import orjson
 
 from . import BridgeDisconnectedError
 from .messages import Response
@@ -68,9 +69,9 @@ class LeapProtocol:
         future.add_done_callback(clean_up)
 
         try:
-            text = json.dumps(cmd)
+            text = orjson.dumps(cmd)
             _LOG.debug("sending %s", text)
-            self._writer.write(text + b"\r\n")
+            self._writer.writelines((text, b"\r\n"))
 
             return await future
         finally:
@@ -84,7 +85,7 @@ class LeapProtocol:
             if received == b"":
                 break
 
-            resp_json = json.loads(received.decode("UTF-8"))
+            resp_json = orjson.loads(received.decode("UTF-8"))
 
             if isinstance(resp_json, dict):
                 tag = resp_json.get("Header", {}).pop("ClientTag", None)
--- a/src/pylutron_caseta/leap.py
+++ b/src/pylutron_caseta/leap.py
@@ -1,7 +1,7 @@
 """LEAP protocol layer."""
 
 import asyncio
-import json
+import orjson as json
 import logging
 import re
 import uuid
@@ -68,7 +68,7 @@ class LeapProtocol:
         future.add_done_callback(clean_up)
 
         try:
-            text = json.dumps(cmd).encode("UTF-8")
+            text = json.dumps(cmd)
             _LOG.debug("sending %s", text)
             self._writer.write(text + b"\r\n")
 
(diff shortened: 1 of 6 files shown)


src/pylutron_caseta/leap.py
```diff
@@ -1,7 +1,7 @@
 """LEAP protocol layer."""
 
 import asyncio
-import json
+import orjson as json
```
bdraco, line 4: My preference would be to keep this imported as `orjson` instead of `json` since they have slightly different APIs and it can be confusing if you are working below and don't scroll up to realize its not standard lib `json`

src/pylutron_caseta/pairing.py
```diff
@@ -61,7 +61,7 @@ async def async_read_json(self, timeout):
 
     async def async_write_json(self, obj):
         """Write an object."""
-        buffer = f"{json.dumps(obj)}\r\n".encode("ASCII")
+        buffer = json.dumps(obj) + b"\r\n"
         self._writer.write(buffer)
```
bdraco, line 65: I think the payload is small here so it probably doesn't matter but if its larger than 2048 bytes `.writelines()` can be used instead as it does [zero copy writes](https://github.com/python/cpython/pull/31871). Otherwise `json.dumps(obj) + b"\r\n"` will result in `memcpy` to create a new immutable bytes string.

bdraco:
I think this is a good change. Some minor nits above.

src/pylutron_caseta/pairing.py
bdraco: `self._write.writelines((json.dumps(obj), b"\r\n"))` would be a tiny but faster since tuples are a bit cheaper than lists

Just but only if the average resulting bytes is > 2048 in length as otherwise it won't matter and you can ignore this suggestion.

jonoberheide: There are a couple cases in there that makes sense (aka the LeapProtocol.request hotpath, not so much the test cases 😆).

bdraco (approved):

bdraco:
Tested with an RA2 select system 👍

src/pylutron_caseta/leap.py
```diff
@@ -84,7 +85,7 @@ async def run(self):
             if received == b"":
                 break
 
-            resp_json = json.loads(received.decode("UTF-8"))
+            resp_json = orjson.loads(received.decode("UTF-8"))
```
bdraco, line 88: The decode here should not be needed as `orjson.loads` accepts `bytes`

src/pylutron_caseta/pairing.py
```diff
@@ -57,12 +57,12 @@ async def async_read_json(self, timeout):
             return None
 
         LOGGER.debug("received: %s", buffer)
-        return json.loads(buffer.decode("UTF-8"))
+        return orjson.loads(buffer.decode("UTF-8"))
```
bdraco, line 60: https://github.com/gurumitts/pylutron-caseta/pull/175/files#r1842602861

tests/test_leap.py
```diff
@@ -109,7 +109,7 @@ async def test_call(pipe: Pipe):
     """Test basic call and response."""
     task = asyncio.create_task(pipe.leap.request("ReadRequest", "/test"))
 
-    received = json.loads((await pipe.test_reader.readline()).decode("utf-8"))
+    received = orjson.loads((await pipe.test_reader.readline()).decode("utf-8"))
```
bdraco, line 112: https://github.com/gurumitts/pylutron-caseta/pull/175/files#r1842602861

src/pylutron_caseta/leap.py
bdraco: https://github.com/ijl/orjson#deserialize

bdraco: > bytes, bytearray, memoryview, and str input are accepted. If the input exists as a memoryview, bytearray, or bytes object, it is recommended to pass these directly rather than creating an unnecessary str object. That is, orjson.loads(b"{}") instead of orjson.loads(b"{}".decode("utf-8")). This has lower memory usage and lower latency.

jonoberheide: 👍

jonoberheide:
> Tested with an RA2 select system 👍

Your Lutron lights are approaching the speed of light now! 🏎💨

I should note I'm testing on QSX.

mdonoughe (approved):
I don't know if this is going to make a noticeable difference because the CPU running pylutron_caseta should be much faster than the CPU in the bridge, or at least that's the case for Caseta.

bdraco:
Might make a difference for some single board systems, then again the pure Python JSON implementation should be in order of magnitude slower or worse if the hub has implemented it in native code

swails:
The hub uses a compiled language (golang). It'll be pretty fast, even on a much slower CPU.

PR #176, opened 2024-11-14
add a helper function test bridge connectivity

jonoberheide:
Hi there! I'm trying to resolve https://github.com/home-assistant/core/issues/110067

@bdraco suggested having separate timeouts in HA-land for the connection phase vs the full-loading phase of all the leap requests/zones/etc. I'm finding that approach challenging as the _connect() logic is so intertwined in the Smartbridge state machine. Instead, a more simple approach may be to define a helper function that tests connectivity to the bridge, which HA can call to verify connectivity and report any issues back to the user, before the full load which will be given a longer timeout (BRIDGE_TIMEOUT) for large QSX deployments.

### DIFF
--- a/src/pylutron_caseta/smartbridge.py
+++ b/src/pylutron_caseta/smartbridge.py
@@ -95,7 +95,7 @@ class Smartbridge:
         except Exception as e:
             _LOG.warning("Test connection to bridge failed: %s", e)
         return False
- 
+
     async def connect(self):
         """Connect to the bridge."""
         # reset any existing connection state
--- a/src/pylutron_caseta/smartbridge.py
+++ b/src/pylutron_caseta/smartbridge.py
@@ -83,6 +83,19 @@ class Smartbridge:
             and self._login_completed.exception() is None
         )
 
+    async def test_connection(self) -> bool:
+        """Attempt to connect to the bridge."""
+        try:
+            async with asyncio_timeout(CONNECT_TIMEOUT):
+                leap = await self._connect()
+                leap.close()
+            return True
+        except asyncio.TimeoutError:
+            _LOG.warning("Test connection to bridge timed out")
+        except Exception as e:
+            _LOG.warning("Test connection to bridge failed: %s", e)
+        return False
+ 
     async def connect(self):
         """Connect to the bridge."""
         # reset any existing connection state


jonoberheide:
Ping, any thoughts here?

mdonoughe:
I don't understand the purpose of this function. If Home Assistant needs to know that progress is being made, shouldn't the regular connection code be sending some sort of event back to say that it's progressing? This function creates a different connection and immediately closes it, which doesn't mean the next connection will complete in any amount of time.

jonoberheide:
> I don't understand the purpose of this function. If Home Assistant needs to know that progress is being made, shouldn't the regular connection code be sending some sort of event back to say that it's progressing? This function creates a different connection and immediately closes it, which doesn't mean the next connection will complete in any amount of time.

Right now, Smartbridge.connect() does more than connect, it also loads all the devices/zones/etc from the bridge. This can take a very long time on large deployments, exceeding HA's timeout (BRIDGE_TIMEOUT, see https://github.com/home-assistant/core/issues/110067). Simply increasing the timeout in HA was not recommended, since if there was some other connection-level issue, we would not want to wait that long.

So two other options are: (1) this PR here where we add a connection test function to ensure reachability, correct creds, etc, which then gives us confidence to increase the timeout and assume the library taking a long time is actually due to it loading all the zones/devices/etc. Or (2) break library API compatibility and separate out the steps into more discrete Smartbridge.connect() and maybe Smartbridge.load()/Smartbridge.monitor() steps so that, again, we can have a short connection timeout, but a much longer loading timeout.

Agreed (1) is kind of a weird workaround, but avoids the API breakage of (2). Maybe there's a (3) I'm not thinking of... :-)

jonoberheide:
Any thoughts on the previous comment and potential approaches?

bdraco:
Maybe add a future for connect completed like login completed, and pass and arrange to call a callback when connect completes

jonoberheide:
@bdraco That makes sense conceptually, but just looking at the login_completed logic and state machine of Smartbridge, I would be in way over my skis.

Ugh, I feel like this is all a lot of complexity for https://github.com/home-assistant/core/issues/110067 compared to just upping the BRIDGE_TIMEOUT (or having a config option to specify a larger timeout)... :-)

bdraco:
> @bdraco That makes sense conceptually, but just looking at the login_completed logic and state machine of Smartbridge, I would be in way over my skis.
> 
> 
> 
> Ugh, I feel like this is all a lot of complexity for https://github.com/home-assistant/core/issues/110067 compared to just upping the BRIDGE_TIMEOUT (or having a config option to specify a larger timeout)... :-) 

It shouldn't be that difficult to pass a callable and save it in the object than call it right after successful connection

Also it looks like this library needs to be updated for Python 3.11 to avoid swallowing cancellation 

https://github.com/gurumitts/pylutron-caseta/blob/ab8006c42393683a200b910625f8cf4ab7b7e4cf/src/pylutron_caseta/smartbridge.py#L853

https://superfastpython.com/asyncio-task-cancellation-best-practices/

bdraco:
> Also it looks like this library needs to be updated for Python 3.11 to avoid swallowing cancellation
> 
> https://github.com/gurumitts/pylutron-caseta/blob/ab8006c42393683a200b910625f8cf4ab7b7e4cf/src/pylutron_caseta/smartbridge.py#L853
> 
> [superfastpython.com/asyncio-task-cancellation-best-practices](https://superfastpython.com/asyncio-task-cancellation-best-practices/)

https://github.com/gurumitts/pylutron-caseta/pull/182

bdraco:
> It shouldn't be that difficult to pass a callable and save it in the object than call it right after successful connection

https://github.com/gurumitts/pylutron-caseta/pull/183

jonoberheide:
Well, you made that look easy! Thanks @bdraco! 🙏

jonoberheide:
Superseded by https://github.com/gurumitts/pylutron-caseta/pull/183

PR #182, opened 2025-02-19
Fix cancellation being swallowed in tasks

bdraco:
`asyncio.CancelledError` was being consumed in the task which means `task.cancelled()` can never be true

https://superfastpython.com/asyncio-task-cancellation-best-practices/#Consume_a_CancelledError_Inside_The_Task

Also nothing in this lib ever cares if the task is actually cancelled or not so its fine to close this PR as its only a problem if in the future something would check or care.

### DIFF
--- a/src/pylutron_caseta/smartbridge.py
+++ b/src/pylutron_caseta/smartbridge.py
@@ -611,13 +611,6 @@ class Smartbridge:
         try:
             while True:
                 await self._monitor_once()
-        except asyncio.CancelledError:
-            if (
-                sys.version_info >= (3, 11)
-                and (task := asyncio.current_task())
-                and task.cancelling()
-            ):
-                raise
         except Exception as ex:
             _LOG.critical("monitor loop has exited", exc_info=1)
             if not self._login_completed.done():
@@ -862,13 +855,6 @@ class Smartbridge:
 
             if not self._login_completed.done():
                 self._login_completed.set_result(None)
-        except asyncio.CancelledError:
-            if (
-                sys.version_info >= (3, 11)
-                and (task := asyncio.current_task())
-                and task.cancelling()
-            ):
-                raise
         except Exception as ex:
             if not self._login_completed.done():
                 self._login_completed.set_exception(ex)
@@ -883,13 +869,6 @@ class Smartbridge:
         except asyncio.TimeoutError:
             _LOG.warning("ping was not answered. closing connection.")
             self._leap.close()
-        except asyncio.CancelledError:
-            if (
-                sys.version_info >= (3, 11)
-                and (task := asyncio.current_task())
-                and task.cancelling()
-            ):
-                raise
         except Exception:
             _LOG.warning("ping failed. closing connection.", exc_info=1)
             self._leap.close()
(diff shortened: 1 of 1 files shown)


mdonoughe:
This way of handling cancellation exceptions seems very un-pythonic.

`SmartBridge` cancels these internal tasks and discards them. If the tasks reraise the cancellation exception it will be a never-retrieved exception.

If the exception is caught and reraised like this instead of being ignored at the bottom of a task stack, nobody can observe it except for the never retrieved exception handler. If the cancellation is supposed to be the result of task returned from the connect function, it needs to be propagated like `Exception` instead of reraised.

bdraco:
> If the tasks reraise the cancellation exception it will be a never-retrieved exception.

I was under the impression that calling `.cancel()` turns off `log_traceback`

https://github.com/python/cpython/blob/0ff16115741aeaaaf7f963f68d5c575efb960277/Lib/asyncio/tasks.py#L205
https://github.com/python/cpython/blob/0ff16115741aeaaaf7f963f68d5c575efb960277/Lib/asyncio/futures.py#L95

bdraco:
If cancellation is swallowed inside the task, `task.cancelled()` will never be true

Example
```python
import asyncio
import gc
import logging

logging.basicConfig(level=logging.DEBUG)


async def test_task_that_swallowed_cancel():
    try:
        await asyncio.Event().wait()
    except asyncio.CancelledError:
        pass


async def swallow_test():
    task = asyncio.create_task(test_task_that_swallowed_cancel())
    await asyncio.sleep(0)
    task.cancel()
    assert task.cancelling()
    await asyncio.sleep(0)
    # cancelled will never be true because cancellation
    # was swallowed
    assert not task.cancelled()
    del task
    gc.collect()
    # Should not log future never retrieved


async def test_task_that_raises_cancel():
    try:
        await asyncio.Event().wait()
    except asyncio.CancelledError:
        raise


async def not_swallow_test():
    task = asyncio.create_task(test_task_that_raises_cancel())
    await asyncio.sleep(0)
    task.cancel()
    assert task.cancelling()
    await asyncio.sleep(0)
    # cancelled will never be true because cancellation
    # was swallowed
    assert task.cancelled()
    del task
    gc.collect()
    # Should not log future never retrieved


asyncio.run(swallow_test())
asyncio.run(not_swallow_test())

```

bdraco:
I think this is a good description of the issue https://superfastpython.com/asyncio-task-cancellation-best-practices/#Consume_a_CancelledError_Inside_The_Task

bdraco:
Also nothing ever cares if the task is actually cancelled or not so its fine to close this PR as its only a problem if in the future something would check or care.

src/pylutron_caseta/smartbridge.py
```diff
@@ -5,6 +5,7 @@
 import math
 import socket
 import ssl
+import sys
```
bdraco, line 8:
```suggestion
```

bdraco:
@mdonoughe Thanks for looking at the PRs.  I'm fine with closing this one out as I don't think its strictly needed.  I'm much more concerned with getting a release done so we can address  @jonoberheide 's [issue](https://github.com/home-assistant/core/issues/110067), which I think also means https://github.com/gurumitts/pylutron-caseta/pull/176 can be closed out as well.

PR #187, opened 2025-04-22
use asyncio.run()

gstein:
There is no event loop to "get", so the example fails.

Instead, use `asyncio.run()` to construct a loop, call the example function, and exit when `example()` completes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,10 +61,8 @@ async def example():
     await bridge.close()
 
 
-# Because pylutron_caseta uses asyncio,
-# it must be run within the context of an asyncio event loop.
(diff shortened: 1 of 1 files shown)


mdonoughe:
> There is no event loop to "get", so the example fails.

That's not how `get_event_loop` works. Despite it's name, it creates a loop if there isn't one already. This was the way to run async code until Python 3.7.

mdonoughe (approved):

gstein:
Ah! Fair point, re: `get_event_loop`. Thanks for the merge!

PR #188, opened 2025-07-13
Adds a new button status to detect double press

rlopezdiez:
Some Keypads, like Alisse, send a MultiTap event every time the user double presses a button. The added const makes it explicit, so integrators like Home Assistant can use it.


rlopezdiez:
Would you mind generating a new version with these changes, please?

mdonoughe:
https://pypi.org/project/pylutron-caseta/0.25.0/

PR #194, opened 2025-11-10
Update Smart Away Subscription

omrishiv:
@mdonoughe sorry for the bonus PR, but I am now able to toggle Smart Away in Home Assistant. Hoping to get your opinion on this/some other changes. 

I had issues here unless I removed the state from this commit.
https://github.com/home-assistant/core/blob/dev/homeassistant/components/lutron_caseta/entity.py#L66-L72

I'm wondering if it's worth changing the state to closer align with devices, using `0` as Smart Away off and `1` as on? I had to put a few conditionals in the component code as this isn't aligned with any other device. Is it worth trying to make this look like a device in the pylutron codebase?


omrishiv:
You're right, thank you. I have opened https://github.com/home-assistant/core/pull/156711

PR #199, opened 2026-04-28
Add zone tuning trim controls (high/low end trim)

r-teller:
## Summary

- Add `set_zone_high_end_trim`, `set_zone_low_end_trim`, and `read_zone_tuning` methods on `Smartbridge` for adjusting the per-zone `HighEndTrim` / `LowEndTrim` values exposed by the LEAP `/zone/{id}/tuningsettings` endpoint.
- Restrict these methods to dimmer types that actually support the endpoint (`WallDimmer`, `DivaSmartDimmer`) via a new `TRIM_SUPPORTED_DEVICE_TYPES` allowlist. Plug-in dimmers and color/white-tune lamps (Ketra/Lumaris) return errors from the bridge for this endpoint and are rejected client-side with `ValueError`.
- The high/low setters read the current tuning settings before writing and refuse a write that would put `HighEndTrim` below `LowEndTrim` (or vice versa). Empirical testing showed the bridge silently accepts these inverted values, so the check is enforced client-side.

`EnergyTrim` is surfaced in the dict returned by `read_zone_tuning` (since the bridge returns it) but no setter is provided — adjusting it has no observable effect on tested hardware.

## Test plan

- [x] `pytest tests/` — 65/65 pass, including new tests covering: success path for each setter, ValueError on unsupported device type (fan controller), ValueError on out-of-range value, ValueError on cross-field violation (HighEndTrim < current LowEndTrim and LowEndTrim > current HighEndTrim), and `read_zone_tuning` parsing.
- [x] `ruff format --check src tests` and `ruff check src tests` — clean.
- [x] Manually verified against a live Caseta bridge using the example script in the PR description: `WallDimmer` zones accept and reflect HighEndTrim updates; `PlugInDimmer`-backed lamps return `BridgeResponseError` on the underlying endpoint, matching the new allowlist behavior.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/pylutron_caseta/smartbridge.py
+++ b/src/pylutron_caseta/smartbridge.py
@@ -22,6 +22,7 @@ from . import (
     FAN_OFF,
     OCCUPANCY_GROUP_UNKNOWN,
     RA3_OCCUPANCY_SENSOR_DEVICE_TYPES,
+    TRIM_SUPPORTED_DEVICE_TYPES,
     BridgeDisconnectedError,
     BridgeResponseError,
 )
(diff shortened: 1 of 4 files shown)


src/pylutron_caseta/smartbridge.py
```diff
@@ -280,6 +281,80 @@ async def get_battery_status(self, device_id: str) -> Optional[str]:
...
+        device = self.devices[device_id]
+        if device.get("type") not in TRIM_SUPPORTED_DEVICE_TYPES:
+            raise ValueError(
+                f"Device {device_id} (type {device.get('type')}) "
+                "does not support tuning settings"
+            )
+        zone_id = device.get("zone")
+        if not zone_id:
+            raise ValueError(f"Device {device_id} has no associated zone")
+        return zone_id
```
mdonoughe, line 300: What is this method for? Does it really take a method to get the zone ID for a device? If the methods act on zones, can they take zone IDs instead of device IDs? There are other methods that take zone IDs.

src/pylutron_caseta/smartbridge.py
```diff
@@ -280,6 +281,80 @@ async def get_battery_status(self, device_id: str) -> Optional[str]:
...
+        :raises KeyError: if the device id is unknown
+        :raises ValueError: if the device type does not support tuning settings
+            or has no associated zone
+        """
+        device = self.devices[device_id]
+        if device.get("type") not in TRIM_SUPPORTED_DEVICE_TYPES:
+            raise ValueError(
+                f"Device {device_id} (type {device.get('type')}) "
+                "does not support tuning settings"
+            )
```
mdonoughe, line 296: This seems like an unnecessary maintenance problem. Why does this library need to validate that the device type supports trim settings?

src/pylutron_caseta/smartbridge.py
```diff
@@ -280,6 +281,80 @@ async def get_battery_status(self, device_id: str) -> Optional[str]:
...
+        self, device_id: str, value: Union[int, float]
+    ) -> None:
+        """Set the high-end trim (0-100) for a dimmable device's zone.
+
+        Performs a read of the current TuningSettings to ensure the new
+        HighEndTrim is not below the current LowEndTrim, since the bridge
+        does not enforce this constraint.
+        """
+        if not 0 <= value <= 100:
+            raise ValueError("HighEndTrim must be 0-100")
```
mdonoughe, line 319: This isn't done for the other methods and probably doesn't need to be done. What happens if the caller tries to set the value outside of the supported range for a device?

src/pylutron_caseta/smartbridge.py
```diff
@@ -280,6 +281,80 @@ async def get_battery_status(self, device_id: str) -> Optional[str]:
...
+        zone_id = self._get_trim_zone_id(device_id)
+        current = await self._read_zone_tuning(zone_id)
+        low = current.get("LowEndTrim")
+        if low is not None and value < low:
+            raise ValueError(f"HighEndTrim {value} would be below LowEndTrim {low}")
```
mdonoughe, line 324: Is this worth it? What happens if high is set less than low? It seems better to have less logic in this library.

tests/test_smartbridge.py
```diff
@@ -1476,6 +1476,184 @@ async def test_set_tilt(bridge: Bridge):
...
+                "EnergyTrim": energy,
+                "LowEndTrim": low,
+            }
+        },
+    )
+
+
+@pytest.mark.asyncio
+async def test_set_zone_high_end_trim(bridge: Bridge):
+    """Test that setting the high-end trim reads then writes."""
```
mdonoughe, line 1500: If this description of the test were accurate, the test shouldn't exist. This test actually tests that the high value can be set. That setting the high value also reads the low value is not a required behavior.

PR #201, opened 2026-05-06
Fix Ketra warm-dim color mode stuck after initial state set

### LINKED ISSUE #200: Ketra warm-dim mode: zone status updates not reflected after initial state load
Basically, if you are using ketra lights and home assistant ever sees warm-dim mode enabled on a light, it will forever not report the light color correctly until it is unset and the integration reloaded.  100% reproducible

**Device**: Ketra SpectrumTune fixture via Caseta/LEAP integration
**Device type reported**: DEVICE_TYPE_SPECTRUM_TUNE

**Behavior**:

Set a Ketra light to any color → HA reflects state correctly and updates on changes ✅
Enable warm-dim mode on the same light → HA captures the initial state on integration restart ✅
Change the warm-dim level (from Lutron app or HA) → _device["warm_dim"] and/or _device["color"] do NOT update until integration restart ❌

**Root cause hypothesis**:
The bridge sends warm-dim state changes via a different LEAP message field (likely ColorTuningProperties or a warm-dim-specific zone status variant) that _handle_one_zone_status in smartbridge.py doesn't parse into device["warm_dim"]. The initial state is populated correctly at startup via the full zone status read, but unsolicited push updates for warm-dim changes are silently dropped.

**To reproduce**:

1. Have a Ketra SpectrumTune fixture connected via LEAP
2. Enable warm-dim mode
3. Subscribe to zone status changes and observe that brightness/color updates in warm-dim mode don't trigger device dict updates

**Expected**: device["warm_dim"] and/or device["color"] should update on every zone status push, including those triggered by warm-dim changes.

timcampos:
## Problem

When a Ketra SpectrumTune fixture is in warm-dim mode, Home Assistant correctly captures the initial state on integration start. However, subsequent color changes (from either the Lutron app or HA) are not reflected in Home Assistant until the integration is restarted.

## Root Cause

When the bridge sends a `ColorTuningStatus` update containing `HSVTuningLevel` (or `XYTuningLevel`) but no `CurveDimming` field, `get_warm_dim_from_leap` returns `None` — meaning "no warm-dim information, leave the flag alone."

This is correct when `ColorTuningStatus` is entirely absent. However, when `ColorTuningStatus` *is* present but simply lacks `CurveDimming`, it means the fixture is actively in a non-warm-dim color mode and the flag should be cleared.

Because `device["warm_dim"]` was never cleared, `color_mode` kept returning `ColorMode.WHITE` indefinitely, masking the correctly-updated `FullColorValue` color state.

## Fix

In `get_warm_dim_from_leap`: when `ColorTuningStatus` is present but `CurveDimming` is absent, return `False` (warm-dim explicitly off) instead of `None` (no opinion). One-line behavioral change.

## Testing

Verified on Ketra SpectrumTune fixtures via RA3/QSX bridge with the Home Assistant lutron_caseta integration. Color state now updates correctly when switching out of warm-dim mode without requiring an integration restart.

Fixes #200

### DIFF
--- a/src/pylutron_caseta/color_value.py
+++ b/src/pylutron_caseta/color_value.py
@@ -108,10 +108,8 @@ class WarmDimmingColorValue:
         Check whether warm dimming is active for a zone status.
 
         Returns None if the zone status does not contain warm dimming information.
-        Returns False if ColorTuningStatus is present but CurveDimming is absent,
-        meaning the bridge has sent a non-warm-dim color update (e.g. HSVTuningLevel
-        from a Ketra fixture) and warm-dim mode should be cleared.
-        Returns True if CurveDimming is present and contains a Curve entry.
+        Returns False if warm diming is inactive
+        Returns True if warm diming is active
         """
         if zone_status is None:
             return None
@@ -122,7 +120,6 @@ class WarmDimmingColorValue:
 
         curve_dimming = color_status.get("CurveDimming")
         if curve_dimming is None:
-            # ColorTuningStatus is present but no CurveDimming — warm-dim is off
             return False
 
         return "Curve" in curve_dimming
--- a/src/pylutron_caseta/color_value.py
+++ b/src/pylutron_caseta/color_value.py
@@ -108,6 +108,10 @@ class WarmDimmingColorValue:
         Check whether warm dimming is active for a zone status.
 
         Returns None if the zone status does not contain warm dimming information.
+        Returns False if ColorTuningStatus is present but CurveDimming is absent,
+        meaning the bridge has sent a non-warm-dim color update (e.g. HSVTuningLevel
+        from a Ketra fixture) and warm-dim mode should be cleared.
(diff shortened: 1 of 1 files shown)


src/pylutron_caseta/color_value.py
```diff
@@ -108,6 +108,10 @@ def get_warm_dim_from_leap(zone_status: dict) -> Optional["bool"]:
         Check whether warm dimming is active for a zone status.
 
         Returns None if the zone status does not contain warm dimming information.
+        Returns False if ColorTuningStatus is present but CurveDimming is absent,
+        meaning the bridge has sent a non-warm-dim color update (e.g. HSVTuningLevel
+        from a Ketra fixture) and warm-dim mode should be cleared.
+        Returns True if CurveDimming is present and contains a Curve entry.
```
mdonoughe: I don't think these details are relevant to the caller. The function returns false if warm dimming is inactive.

mdonoughe (approved):

PR #202, opened 2026-05-06
Add support for Vierti Hybrid Keypad (RadioRA 3)

albanh:
## Summary

The Vierti Hybrid Keypad is a newer RadioRA 3 device that combines a dimmer with keypad buttons (models HD-4B and HD-3RL). The LEAP API reports these with an obfuscated device type string `N-CVLpqtTTOIeG-4hcjKgw` rather than a human-readable name like other keypads (e.g. `SunnataKeypad`).

Because this device type is not in `_LEAP_DEVICE_TYPES["sensor"]`, the `_load_ra3_station_device` method silently skips these devices, meaning no button entities are created and no button events are fired.

## Changes

- Added `N-CVLpqtTTOIeG-4hcjKgw` to the `sensor` list in `_LEAP_DEVICE_TYPES`
- Added a friendly name mapping (`ViertiKeypad`) in `_load_ra3_station_device` since the obfuscated string doesn't match the existing `"Keypad" in device_type` check

## Environment

- RadioRA 3 processor running LEAP v3.247
- Vierti models: HD-4B (4-button hybrid) and HD-3RL (3-button with raise/lower hybrid)
- The dimmer zones for these devices already work correctly via `_load_ra3_zones`; only the keypad button discovery was missing

## LEAP API data

The device type was identified by querying `/area/{id}/associatedcontrolstation` on the RA3 processor:

```json
{
  "Device": {
    "href": "/device/558",
    "DeviceType": "N-CVLpqtTTOIeG-4hcjKgw",
    "AddressedState": "Addressed"
  }
}
```

Individual device query (`/device/558`) confirmed:

```json
{
  "ModelNumber": "HD-4B",
  "SerialNumber": 128610831,
  "LocalZones": [{"href": "/zone/589"}],
  "LinkNodes": [{"href": "/device/558/linknode/560"}]
}
```

### DIFF
--- a/src/pylutron_caseta/smartbridge.py
+++ b/src/pylutron_caseta/smartbridge.py
@@ -1092,6 +1092,8 @@ class Smartbridge:
             device_type_friendly = "Pico"
         elif "Keypad" in device_type:
             device_type_friendly = "Keypad"
+        elif device_type == "N-CVLpqtTTOIeG-4hcjKgw":
+            device_type_friendly = "ViertiKeypad"
 
         if "SerialNumber" in device_json.Body["Device"]:
             device_serial = device_json.Body["Device"]["SerialNumber"]
--- a/src/pylutron_caseta/smartbridge.py
+++ b/src/pylutron_caseta/smartbridge.py
@@ -1093,7 +1093,7 @@ class Smartbridge:
         elif "Keypad" in device_type:
             device_type_friendly = "Keypad"
         elif device_type == "N-CVLpqtTTOIeG-4hcjKgw":
-            device_type_friendly = "ViertiKeypad"
+            device_type_friendly = "Keypad"
 
         if "SerialNumber" in device_json.Body["Device"]:
             device_serial = device_json.Body["Device"]["SerialNumber"]
--- a/src/pylutron_caseta/smartbridge.py
+++ b/src/pylutron_caseta/smartbridge.py
@@ -17,7 +17,9 @@ except ImportError:
     from asyncio import get_event_loop as get_loop
 
 from . import (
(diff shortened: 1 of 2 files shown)


src/pylutron_caseta/smartbridge.py
```diff
@@ -1092,6 +1092,8 @@ async def _load_ra3_station_device(
             device_type_friendly = "Pico"
         elif "Keypad" in device_type:
             device_type_friendly = "Keypad"
+        elif device_type == "N-CVLpqtTTOIeG-4hcjKgw":
+            device_type_friendly = "ViertiKeypad"
```
mdonoughe: Shouldn't this just be "Keypad" like all the other keypads?

mdonoughe (commented):
Does the dimmer functionality work with this patch? Is it finally time to stop using these device type mappings and use the capabilities returned by the bridge?

src/pylutron_caseta/smartbridge.py
albanh: Good point. I used `ViertiKeypad` to be explicit about which device it mapped to, but there's no functional reason for it to differ from the other keypads. I'll change it to `Keypad`.

albanh:
Yes, dimmer functionality continues to work with this patch. The dimmer zones are handled by `_load_ra3_zones` which is independent of the station/keypad discovery path this PR touches.

On moving to a capabilities-based approach: I agree that would be a much better long-term solution, especially as more devices show up with obfuscated type strings. This PR is just a targeted fix to unblock a specific device, but I'd be happy to look into a capabilities-based refactor as a separate effort if that's something the project wants to pursue.

PR #205, opened 2026-06-20
Add Natural Light Optimization support

potatosalad:
## Summary

Adds first-class Smartbridge APIs for Lutron Natural Light Optimization:

- discover configured NLO groups
- read and subscribe to NLO enabled/disabled status
- enable or disable an NLO group
- gracefully ignore unsupported NLO endpoints on processors without the feature

This lets consumers use NLO without reaching into private LEAP request helpers.

See https://support.lutron.com/us/en/product/homeworks/article/lutron-designer/Natural-Light-Optimization-expansion-to-include-Roller-Shades

image

## Testing

Tests passed.

### DIFF
--- a/src/pylutron_caseta/smartbridge.py
+++ b/src/pylutron_caseta/smartbridge.py
@@ -277,27 +277,37 @@ class Smartbridge:
         """Will return all known scenes from the Smart Bridge."""
         return self.scenes
 
-    async def get_natural_light_optimizations(self) -> Dict[str, dict]:
-        """Read and return all known Natural Light Optimizations."""
-        self.natural_light_optimizations = {}
+    def get_natural_light_optimizations(self) -> Dict[str, dict]:
+        """Will return all known Natural Light Optimizations."""
+        return self.natural_light_optimizations
+
+    async def refresh_natural_light_optimizations(self) -> Dict[str, dict]:
+        """Read Natural Light Optimizations from the bridge."""
+        natural_light_optimizations: Dict[str, dict] = {}
         try:
             natural_light_optimization_json = await self._request(
                 "ReadRequest", "/system/naturallightoptimization"
             )
         except BridgeResponseError as ex:
             if _bridge_error_is_not_supported(ex):
+                self.natural_light_optimizations = natural_light_optimizations
                 return self.natural_light_optimizations
             raise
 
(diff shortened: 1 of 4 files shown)


src/pylutron_caseta/smartbridge.py
```diff
@@ -262,6 +277,30 @@ def get_scenes(self) -> Dict[str, dict]:
         """Will return all known scenes from the Smart Bridge."""
         return self.scenes
 
+    async def get_natural_light_optimizations(self) -> Dict[str, dict]:
+        """Read and return all known Natural Light Optimizations."""
+        self.natural_light_optimizations = {}
```
mdonoughe: Doesn't setting `self.natural_light_optimizations = {}` at the start of this method mean that:
1. If the update fails for some reason, `self.natural_light_optimizations` will remain empty until the next time this function is called and completes successfully?
2. If the update fails in the for loop, some but not all of the of the values will be present, but without their statuses?
3. Even if everything is working correctly, the set of optimizations will briefly become empty while this function is executing?

src/pylutron_caseta/smartbridge.py
```diff
@@ -262,6 +277,30 @@ def get_scenes(self) -> Dict[str, dict]:
...
+        try:
+            natural_light_optimization_json = await self._request(
+                "ReadRequest", "/system/naturallightoptimization"
+            )
```
mdonoughe, line 290: I think all the other "get" methods just get something from an instance variable, but this one makes a request to the bridge. Maybe that's necessary if the bridge doesn't notify LEAP clients of changes in the available optimizations. If the method always retrieves the optimizations from the bridge on demand, is `self.natural_light_optimizations` necessary?

src/pylutron_caseta/smartbridge.py
potatosalad: Yup, I missed that on the first pass.  I split things out into a separate function `refresh_natural_light_optimizations()` which only swaps out the instance variable if it's successful, otherwise it leaves the instance variable alone.  What do you think?

src/pylutron_caseta/smartbridge.py
potatosalad: Split things out to a separate `refresh_natural_light_optimizations()` function, does that line up more with what you were thinking?

PR #207, opened 2026-07-07
Close connection when re-login fails so login is retried

### LINKED ISSUE #206: Bridge left "half-alive" (commands work, no status updates) when re-login fails after reconnect — e.g. after a processor reboot
Environment: pylutron-caseta 0.28.0, HomeWorks QSX (HQP7-2), via Home Assistant 2026.7.1 lutron_caseta.

**Symptoms**
1. Processor rebooted (firmware install). Expected `Reconnecting after error: [Errno 111] Connect call failed` until it came back.
2. Once it accepted connections again: `pylutron_caseta.leap` ERROR `Was not expecting message with tag <uuid>: {... 'Url': '/device?where=IsThisDevice:true' ...}`.
3. From then on the session is half-alive: CreateRequest zone commands round-trip fine (201 Created, lights actually change), but ALL zone-status/button subscription updates stop forever. Home Assistant entities freeze. No further errors.

**Root cause**
A processor that has just rebooted accepts TCP but answers slowly. During re-login, a request (here `/device?where=IsThisDevice:true` in `_load_ra3_processor`) exceeds `REQUEST_TIMEOUT` (5 s). The timeout pops the tag from `LeapProtocol._in_flight_requests`, so the late reply triggers the "Was not expecting message" ERROR (harmless in itself — the message is dropped).

The real problem is in `Smartbridge._login`'s exception handler: on a *re*connect, `_login_completed` is already done (from the first login and intentionally not reset), so the exception is set nowhere and re-raised into a task nobody awaits. The login task dies silently, mid-sequence. Subscriptions live in the per-connection `LeapProtocol._tagged_subscriptions`, which is a fresh empty dict on the new connection, and they are only re-registered by `_login()`. Result: connection healthy, requests and ping fine, `logged_in` still True (stale `_login_completed`), but no SubscribeRequest was ever sent — status updates never resume.

**Reproduction:** reboot the bridge/processor mid-session (or otherwise make one re-login request exceed 5 s while the connection stays up). Deterministic repro in the regression test included with the PR: fail the first re-login request with a timeout.

**Suggested fix:** if `_login()` fails after the initial login already completed, close the LEAP connection (same pattern `_ping()` uses), so `_monitor` reconnects and login retries to completion. PR attached.

bayan9:
Fixes #206

After a reconnect, `_login()` runs as a fire-and-forget task. If it raises (e.g. a request times out against a processor that is still booting), the failure is silent: `_login_completed` is already resolved from the initial session, nothing awaits `_login_task`, and the connection stays open. Requests keep working, but the status subscriptions — which live in the new connection's `LeapProtocol._tagged_subscriptions` and are only registered by `_login()` — are never re-established. The bridge is then permanently "half-alive": commands succeed, status updates never arrive.

This change makes a failed re-login close the connection (mirroring `_ping()`'s existing failure handling). `LeapProtocol.run()` then ends, `_monitor` reconnects after `RECONNECT_DELAY`, and login retries until it completes, subscriptions included. Initial-connect behavior is unchanged (`connect()` still surfaces the exception via `_login_completed`).

Adds `test_relogin_failure_closes_connection`, following the existing `test_reconnect_*` fake-time/fake-LEAP conventions in `tests/test_smartbridge.py`. It hangs (fails on timeout) without the fix and passes with it. Full suite: 60 passed.

### DIFF
--- a/tests/test_smartbridge.py
+++ b/tests/test_smartbridge.py
@@ -1644,6 +1644,47 @@ async def test_reconnect_error(bridge: Bridge):
 
 
 @pytest.mark.asyncio
+async def test_relogin_failure_closes_connection(bridge: Bridge):
+    """Test that a failed re-login tears down the connection and retries.
+
+    Regression test for the "half-alive" state seen when a processor reboots:
+    if a request made during re-login fails (for example, it times out because
+    the processor is still starting up), the bridge must close the connection
+    and retry login on a new connection instead of leaving a session where
+    commands work but status subscriptions were never re-established.
+    """
+    time = 0.0
+    asyncio.get_running_loop().time = lambda: time  # type: ignore [method-assign]
+
+    bridge.disconnect()
+
+    await asyncio.sleep(0.0)
+    time += smartbridge.RECONNECT_DELAY
+
+    # Accept the new connection, but fail the first login request, simulating
+    # a processor that accepts connections but is too slow to answer.
+    leap = await bridge.connections.get()
+    request, response = await leap.requests.get()
+    assert request == Request(communique_type="ReadRequest", url="/area")
+    response.set_exception(asyncio.TimeoutError())
+    leap.requests.task_done()
+    bridge.connections.task_done()
+
+    # The bridge should close the half-logged-in connection and reconnect.
+    # accept_connection completes the full login sequence, including
+    # re-subscribing to status updates.
+    # Yield until the monitor loop has observed the closed connection and
+    # scheduled its reconnect delay, then advance the clock past the delay.
+    for _ in range(10):
+        await asyncio.sleep(0.0)
+    time += smartbridge.RECONNECT_DELAY
+
+    await bridge.accept_connection()
+
+    assert bridge.target.is_connected()
+
+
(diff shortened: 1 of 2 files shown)


tests/test_smartbridge.py
```diff
@@ -1643,6 +1643,47 @@ async def test_reconnect_error(bridge: Bridge):
...
+    leap.requests.task_done()
+    bridge.connections.task_done()
+
+    # The bridge should close the half-logged-in connection and reconnect.
+    # accept_connection completes the full login sequence, including
+    # re-subscribing to status updates.
+    # Yield until the monitor loop has observed the closed connection and
+    # scheduled its reconnect delay, then advance the clock past the delay.
+    for _ in range(10):
+        await asyncio.sleep(0.0)
```
mdonoughe, line 1679: This doesn't seem very reliable. We probably should set the delay to 0 for this test? Otherwise we probably want a robust mock time implementation than yielding a set number of times. I'm sure there are libraries for it, but maybe it's enough to just modify `asyncio.get_running_loop().call_later` to keep track of the expected wake up times.

PR #208, opened 2026-07-17
Add OpenCloseStop cover discovery

alexgorbachev:
## Summary

- discover HomeWorks QSX zones reporting `ControlType: OpenCloseStop` as covers
- route Raise, Lower, and Stop through the standard LEAP zone command endpoints
- keep state unknown because these zones do not report a position
- document the observed notification limits and future extension point

## Applicability

Discovery is based on the Lutron control type, not on a processor model, project-specific zone ID, or room name. It should therefore apply to installations where QSX enumerates a zone as `OpenCloseStop` and accepts the standard Raise, Lower, and Stop commands.

The sanitized fixture also covers an identity-only zone-status entry. The notifications observed from these zones did not contain level, direction, or motion attributes, so this change deliberately preserves `current_state == -1` after discovery and commands instead of treating the device like a position-capable shade. If another device or future Lutron firmware supplies additional raw status attributes, direction or position support can be added separately without changing this discovery behavior.

## Validation

- test suite passes on Python 3.10, 3.11, 3.12, 3.13, and 3.14
- Python 3.14 Ruff formatting/checks and mypy pass
- wheel and source distribution build successfully as version `0.29.0`
- exact LEAP Raise, Lower, and Stop request URLs and bodies are covered

The fixture and documentation contain no processor address, certificate, room name, serial number, or installation-specific zone ID.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,35 @@ async def example():
 asyncio.run(example())
 ```
 
+### Open/close/stop covers
+
+HomeWorks QSX and RadioRA 3 projects expose logical zones through the LEAP
+`/area/{area_id}/associatedzone` resource. `pylutron-caseta` represents each
+zone as a device and uses its `ControlType` for domain classification. A zone
+whose control type is exactly `OpenCloseStop` is returned by
+`get_devices_by_domain("cover")`, regardless of its zone ID, area name, load
+name, or physical motor controller.
+
+This supports any HomeWorks QSX or RadioRA 3 project that reports the same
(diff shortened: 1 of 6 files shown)


mdonoughe (approved):

ehoch:
Thank you for this, @alexgorbachev! This actually got some of my shades that are powered by relay modules to work in HA! @mdonoughe do you know when 0.30 will be released? I need to make a PR to core as well to get this fully working.

PR #209, opened 2026-07-26
Fix button subscriber lookup: convert int button_id to str

sarielhp:
## Description

`_handle_button_status` uses `button_id` (an integer from `id_from_href()`) to look up callbacks in `_button_subscribers`, but `add_button_subscriber` registers them with `str(button_id)` (a string). The type mismatch causes every button event lookup to fail, so no subscribers are ever notified.

## Fix

Convert `button_id` to `str(button_id)` before both the membership test and the callback invocation in `_handle_button_status`.

## Testing

After this fix, button press/release events from Pico remotes are correctly delivered to registered subscribers (e.g., the Home Assistant `lutron_caseta` integration's `_async_button_event` callback), and automations triggered by `lutron_caseta_button_event` work as expected.

### DIFF
--- a/src/pylutron_caseta/smartbridge.py
+++ b/src/pylutron_caseta/smartbridge.py
@@ -767,9 +767,12 @@ class Smartbridge:
         if button_id in self.buttons:
             self.buttons[button_id]["current_state"] = button_event
             # Notify any subscribers of the change to button status
-            if button_id in self._button_subscribers:
-                self._button_subscribers[button_id](button_event)
+#            if button_id in self._button_subscribers:
+#                self._button_subscribers[button_id](button_event)
+            if str(button_id) in self._button_subscribers:
+                self._button_subscribers[str(button_id)](button_event)
 
+                
     def _handle_button_led_status(self, response: Response):
         """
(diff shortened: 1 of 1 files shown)


src/pylutron_caseta/smartbridge.py
```diff
@@ -767,9 +767,12 @@ def _handle_button_status(self, response: Response):
         if button_id in self.buttons:
             self.buttons[button_id]["current_state"] = button_event
             # Notify any subscribers of the change to button status
-            if button_id in self._button_subscribers:
-                self._button_subscribers[button_id](button_event)
+#            if button_id in self._button_subscribers:
+#                self._button_subscribers[button_id](button_event)
```
mdonoughe, line 771: The old code shouldn't be left behind as a comment.

mdonoughe (commented):
The premise of this PR is incorrect. `id_from_href()` can only return strings. There is no reason to convert the strings to strings again. The suggested change would not fix the perceived problem anyway, because `self.buttons` is also a dictionary that is keyed by strings.

PR #212, opened 2026-07-30
Add on_disconnect_callback parameter to Smartbridge

brio1337:
## Summary

Adds an optional `on_disconnect_callback` parameter to `Smartbridge`, mirroring the existing `on_connect_callback` pattern. The callback fires once when the connection transitions from connected to disconnected.

## Motivation

Home Assistant's Lutron Caséta integration needs to mark entities as unavailable when the bridge goes offline. Currently there's no way to detect unexpected disconnections without polling `is_connected()`.

Related to Home Assistant's [Integration Quality Scale](https://developers.home-assistant.io/docs/core/integration-quality-scale/rules/entity-unavailable/) which requires entities to be marked unavailable when their device/service is unreachable.

## Design decisions

- **Does NOT fire on intentional `close()`**: Only fires for unexpected disconnections (session ended or connection error)

### DIFF
--- a/src/pylutron_caseta/smartbridge.py
+++ b/src/pylutron_caseta/smartbridge.py
@@ -51,6 +51,7 @@ class Smartbridge:
         self,
         connect: Callable[[], Coroutine[Any, Any, LeapProtocol]],
         on_connect_callback: Optional[Callable[[], None]] = None,
+        on_disconnect_callback: Optional[Callable[[], None]] = None,
     ) -> None:
         """Initialize the Smart Bridge."""
         self.devices: Dict[str, dict] = {}
@@ -75,6 +76,8 @@ class Smartbridge:
         self._monitor_task: Optional[asyncio.Task] = None
         self._ping_task: Optional[asyncio.Task] = None
         self._on_connect_callback = on_connect_callback
+        self._on_disconnect_callback = on_disconnect_callback
+        self._connected = False
 
     @property
     def logged_in(self):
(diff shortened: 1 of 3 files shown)


src/pylutron_caseta/smartbridge.py
```diff
@@ -698,6 +703,9 @@ async def _monitor_once(self):
             self._ping_task = asyncio.get_running_loop().create_task(self._ping())
 
             await self._leap.run()
+            if self._connected and self._on_disconnect_callback:
+                self._on_disconnect_callback()
+            self._connected = False
```
mdonoughe: It's kind of weird that `_connected` is set to false after running the callback. It shouldn't really matter because there's no way for somebody to request the value to notice it's still true while running the disconnect callback, but there is a minor consequence where if the callback throws certain types of exception, the exception is caught and then the disconnect handler is called again because `_connected` is still true.

The logic in this function is not very good. I don't think this new code or the sleep really belong inside this try block. Maybe the thing to do would be to move the callback and the sleep from both the try block and the except block to the end of the function after the finally, leaving only the log warnings behind. Then this code only needs to be written once.

mdonoughe: There is actually an `is_connected` which isn't being updated here. `is_connected` will probably return true during this callback.

brio1337: Agreed, `is_connected` is confusing. After we connect and finish logging in the first time, `is_connected` returns true, and continues returning true even if we lose connection and are in a reconnect loop.

I'm trying to add the ability for the client to be able to see that we've lost connection. I added `_connected` as an internal flag so `on_disconnected_callback` would only be called once per `on_connected_callback`, and I was trying to avoid changing any existing behaviour. But now I'm thinking that adding `on_disconnect_callback` may not be enough if `is_connected` is still true.

Thoughts on changing `is_connected` to reflect the actual connection state?

mdonoughe (approved):

src/pylutron_caseta/smartbridge.py
mdonoughe: It should probably be the actual connection state.

PR #213, opened 2026-08-20
Expose raw zone status events

alexgorbachev:
## Summary

- add an immutable `ZoneStatusEvent` containing the zone ID, resolved device ID, raw status payload, and delivery origin
- add opt-in, multi-subscriber `subscribe_zone_status(callback)` and `unsubscribe_zone_status(callback)` methods
- distinguish snapshots from notifications without implying change detection or deduplication
- publish events after existing device-state processing while preserving attributes that pylutron-caseta does not currently interpret

## Motivation

pylutron-caseta already receives zone-status messages from Lutron processors, but consumers can observe them only indirectly through the normalized device-state subscriber. That interface works well for entities that reread a known property after a notification. It does not provide enough information for consumers that need to:

- observe a notification that contains no recognized state attribute
- distinguish initialization or reconnection snapshots from notifications
- retain attributes introduced by another device type or future Lutron firmware
- allow more than one independent consumer to observe the same event
- remove a subscription cleanly during consumer shutdown

`add_subscriber()` currently stores one callback per device, passes no event payload, provides no delivery-origin metadata, and has no explicit unsubscribe method. Delivering raw events through that callback would change its established contract. Unifying the existing device, button, occupancy, and related subscriber mechanisms would also broaden this change into restructuring existing notification paths, including their ordering and error-handling behavior.

This contribution adds a separate zone-status API using the paired registration/removal pattern already used by `LeapProtocol` unsolicited listeners. Existing subscribers and normalized device-state behavior remain unchanged. The new listeners observe messages Smartbridge already processes; they do not create additional protocol subscriptions. When nobody subscribes, no event payload is copied or constructed.

## API and general applicability

Register and remove a listener using the same callback:

```python
bridge.subscribe_zone_status(handle_zone_status)
# During consumer shutdown:
bridge.unsubscribe_zone_status(handle_zone_status)
```

Each subscribe call adds one registration, including duplicate callbacks. Each unsubscribe call removes one matching registration and raises `ValueError` if none exists, following the low-level unsolicited-listener API.

Each subscriber receives a `ZoneStatusEvent`, for example:

```python
ZoneStatusEvent(
    zone_id="123",
    device_id="123",
    status={"Zone": {"href": "/zone/123"}},
    origin=ZoneStatusEventOrigin.SNAPSHOT,
)
```

- `SNAPSHOT` identifies status obtained from initialization reads or subscription responses, including reconnects.
- `NOTIFICATION` identifies status delivered through a notification callback.

These values describe the delivery path, not whether a value changed, and do not guarantee an ordering between snapshots and notifications. A reconnect snapshot may contain changes made while disconnected. A notification may repeat unchanged values or concern attributes a consumer does not use. Both are processed normally and delivered without deduplication. Neither origin proves that a command or physical movement occurred.

The raw status is copied into recursively immutable containers before delivery. One subscriber cannot modify the event observed by another, and later mutation of the decoded response cannot change an event already delivered. Subscriber exceptions are isolated and logged so they do not interrupt normal state processing or other raw-event subscribers.

The API is based on ordinary Lutron zone-status messages and is not limited to a particular zone type, processor model, room, or integration. It provides a general primitive for maintaining derived state, correlating separately known operations with processor notifications, reconciling or invalidating cached state, diagnostics, and observing attributes that pylutron-caseta does not yet normalize.

The library does not infer direction, position, command source, or intent from the event. Those interpretations remain the consumer's responsibility and must be based on available protocol evidence.

## Home Assistant context

This follows #208, which added discovery and Raise, Lower, and Stop commands for `OpenCloseStop` zones.

The immediate motivation is Home Assistant support for covers whose processor does not report physical position. Home Assistant can maintain an estimated position using calibrated travel time, correlate notifications with separately known actions, and invalidate the estimate after unexplained notifications or reconnects. This does not require treating every notification as proof of movement or ignoring changed state in snapshots.

This PR supplies only the general zone-event primitive. Position estimation, timing, command correlation, and user configuration remain outside pylutron-caseta.

## Compatibility

- existing per-device, button, occupancy, and Smart Away subscribers are unchanged
- normalized device-state mutation and existing device callbacks run before the new event is emitted
- consumers that do not register a zone-status subscriber incur no new event-copying work
- removing one callback registration does not remove other registrations
- raw-event subscriber failures do not prevent delivery to other raw-event subscribers

## Validation

- 67 tests pass on Python 3.10, 3.11, 3.12, 3.13, and 3.14
- Python 3.14 Ruff formatting/checks and mypy pass
- GitHub CI and Coveralls pass on the latest head, `b2c4a8a`
- wheel and source distribution build successfully as version `0.29.0`
- tests cover Caseta, RA3, and QSX reconnects; changed state delivered in reconnect snapshots; repeated identical notifications; multiple subscribers; callback registration and removal; subscriber failure isolation; unknown future attributes; and recursive payload immutability

### DIFF
--- a/src/pylutron_caseta/smartbridge.py
+++ b/src/pylutron_caseta/smartbridge.py
@@ -1,12 +1,15 @@
 """Provides an API to interact with the Lutron Caseta Smart Bridge & RA3 Processor."""
 
 import asyncio
+from copy import deepcopy
+from dataclasses import dataclass
+from enum import Enum
 import logging
 import math
 import socket
 import ssl
 from datetime import timedelta
-from typing import Callable, Dict, List, Optional, Tuple, Union, Coroutine, Any
+from typing import Any, Callable, Coroutine, Dict, List, Mapping, Optional, Tuple, Union
 from .color_value import ColorMode, WarmDimmingColorValue
 
 
@@ -40,6 +43,23 @@ REQUEST_TIMEOUT = 5.0
 RECONNECT_DELAY = 2.0
 
 
+class ZoneStatusEventOrigin(Enum):
+    """Identify whether a zone status is a snapshot or a live update."""
+
+    INITIAL = "initial"
+    UPDATE = "update"
+
+
+@dataclass(frozen=True)
+class ZoneStatusEvent:
+    """Describe a zone status without interpreting device-specific attributes."""
+
+    zone_id: str
+    device_id: str
+    status: Mapping[str, Any]
+    origin: ZoneStatusEventOrigin
+
+
 class Smartbridge:
     """
     A representation of the Lutron Caseta Smart Bridge.
@@ -62,6 +82,7 @@ class Smartbridge:
         self.smart_away_state: str = ""
         self._connect = connect
         self._subscribers: Dict[str, Callable[[], None]] = {}
+        self._zone_status_subscribers: List[Callable[[ZoneStatusEvent], None]] = []
         self._occupancy_subscribers: Dict[str, Callable[[], None]] = {}
         self._button_subscribers: Dict[str, Callable[[str], None]] = {}
         self._smart_away_subscriber: Optional[Callable[[str], None]] = None
@@ -168,6 +189,21 @@ class Smartbridge:
         :param callback_: callback to invoke
         """
         self._subscribers[device_id] = callback_
+
+    def add_zone_status_subscriber(
+        self, callback_: Callable[[ZoneStatusEvent], None]
+    ) -> Callable[[], None]:
+        """Subscribe to raw zone statuses and return an idempotent unsubscribe call."""
+        if not callable(callback_):
+            raise TypeError("callback must be callable")
+
+        self._zone_status_subscribers.append(callback_)
+
+        def unsubscribe() -> None:
+            if callback_ in self._zone_status_subscribers:
+                self._zone_status_subscribers.remove(callback_)
+
+        return unsubscribe
 
     def add_occupancy_subscriber(
         self, occupancy_group_id: str, callback_: Callable[[], None]
@@ -729,9 +765,11 @@ class Smartbridge:
         body = response.Body
         if body is None:
             return
-        self._handle_zone_status(body["ZoneStatus"])
+        self._handle_zone_status(body["ZoneStatus"], ZoneStatusEventOrigin.UPDATE)
 
-    def _handle_zone_status(self, status):
+    def _handle_zone_status(
+        self, status: dict[str, Any], origin: ZoneStatusEventOrigin
+    ) -> None:
         zone = id_from_href(status["Zone"]["href"])
         level = status.get("Level", -1)
         fan_speed = status.get("FanSpeed", None)
@@ -754,6 +792,18 @@ class Smartbridge:
 
         if device["device_id"] in self._subscribers:
             self._subscribers[device["device_id"]]()
+
+        event = ZoneStatusEvent(
+            zone_id=zone,
+            device_id=device["device_id"],
+            status=deepcopy(status),
+            origin=origin,
+        )
+        for callback in tuple(self._zone_status_subscribers):
+            try:
+                callback(event)
+            except Exception:  # pylint: disable=broad-except
+                _LOG.exception("Zone status subscriber raised an exception")
 
     def _handle_button_status(self, response: Response):
         _LOG.debug("Handling button status: %s", response)
@@ -791,14 +841,18 @@ class Smartbridge:
             if button_led_id in self._subscribers:
                 self._subscribers[button_led_id]()
 
-    def _handle_multi_zone_status(self, response: Response):
+    def _handle_multi_zone_status(
+        self,
+        response: Response,
+        origin: ZoneStatusEventOrigin = ZoneStatusEventOrigin.UPDATE,
+    ) -> None:
         _LOG.debug("Handling multi zone status: %s", response)
 
         if response.Body is None:
             return
 
         for zonestatus in response.Body["ZoneStatuses"]:
-            self._handle_zone_status(zonestatus)
+            self._handle_zone_status(zonestatus, origin)
 
     def _handle_occupancy_group_status(self, response: Response):
         _LOG.debug("Handling occupancy group status: %s", response)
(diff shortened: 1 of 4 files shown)


src/pylutron_caseta/smartbridge.py
```diff
@@ -40,6 +44,34 @@
...
+
+def _immutable_copy(value: Any) -> Any:
+    """Copy a JSON-like value into recursively immutable containers."""
+    if isinstance(value, Mapping):
+        return MappingProxyType(
+            {key: _immutable_copy(item) for key, item in value.items()}
+        )
+    if isinstance(value, list):
+        return tuple(_immutable_copy(item) for item in value)
+    return deepcopy(value)
```
mdonoughe, line 72: Why do strings, numbers, booleans, and null need to be deep copied?

src/pylutron_caseta/smartbridge.py
```diff
@@ -169,6 +202,25 @@ def add_subscriber(self, device_id: str, callback_: Callable[[], None]):
         """
         self._subscribers[device_id] = callback_
 
+    def add_zone_status_subscriber(
+        self, callback_: Callable[[ZoneStatusEvent], None]
+    ) -> Callable[[], None]:
+        """Subscribe to raw zone statuses and return an idempotent unsubscribe call."""
```
mdonoughe, line 208: This new function doesn't follow the same subscribe/unsubscribe pattern as the existing subscription functions.

src/pylutron_caseta/smartbridge.py
```diff
@@ -40,6 +44,34 @@
 RECONNECT_DELAY = 2.0
 
 
+class ZoneStatusEventOrigin(Enum):
+    """Identify whether a zone status is a snapshot or a live update."""
```
mdonoughe, line 48: I don't think this is useful.

If the bridge publishes a state change when nothing has changed or when the changed attribute isn't the attribute you care about, you will get a duplicate update event anyway.

If the state changes when the bridge is not connected, you will get a real state change that is marked as initial and whatever you're expecting to do as a result of the state change will not happen.

src/pylutron_caseta/smartbridge.py
```diff
@@ -40,6 +44,34 @@
...
+
+    INITIAL = "initial"
+    UPDATE = "update"
+
+
+@dataclass(frozen=True)
+class ZoneStatusEvent:
+    """Describe a zone status without interpreting device-specific attributes."""
```
mdonoughe, line 56: These new types have function-like documentation.

src/pylutron_caseta/smartbridge.py
```diff
@@ -755,6 +813,21 @@ def _handle_zone_status(self, status):
         if device["device_id"] in self._subscribers:
             self._subscribers[device["device_id"]]()
 
+        if not self._zone_status_subscribers:
+            return
```
mdonoughe, line 817: Returning earlier if there are no zone status subscribers from a function that does things besides notifying zone status subscribers seems like something that will surprise somebody in the future.

mdonoughe, line 817: Should the other code that updates the pylutron_caseta device status also be a zone status subscriber?

tests/test_smartbridge.py
```diff
@@ -647,7 +648,7 @@ def disconnect(self, exception=None):
         else:
             self.leap.running.set_exception(exception)
 
-    async def accept_connection(self):
+    async def accept_connection(self, processor=CASETA_PROCESSOR):
```
mdonoughe, line 651: Telling the bridge what processor it is here seems pretty weird. Should it remember what processor it is when it is initialized? This change fixes a problem where if you disconnect and reconnect a RA3 processor it turns into a Caseta processor, but keeps the ability to change the processor on reconnect and keeps the default of changing to Caseta.

src/pylutron_caseta/smartbridge.py
alexgorbachev: Ah yes. Of course. No need to do that for scalars. Thank you.

src/pylutron_caseta/smartbridge.py
alexgorbachev: You are right. Reworked.

Changed this to paired `subscribe_zone_status(callback)` and `unsubscribe_zone_status(callback)` methods, following the unsolicited-listener API. Unsubscribe removes one matching registration and raises ValueError if none remains.

src/pylutron_caseta/smartbridge.py
alexgorbachev: Good point.
Updated the docstrings to describe the data represented by these types rather than actions they perform.

src/pylutron_caseta/smartbridge.py
alexgorbachev: Makes sense. Replaced the early return with a conditional around event construction and notification only.

src/pylutron_caseta/smartbridge.py
alexgorbachev: Both points are valid. The distinction is the delivery path, not whether a value changed. I’ve renamed the values to SNAPSHOT and NOTIFICATION and clarified the contract.

Both kinds still update device state and reach subscribers without deduplication. A reconnect snapshot may contain real changes, and a notification may repeat unchanged values. Neither proves that a command or physical movement occurred.

The distinction lets consumers reconcile state after reconnect without treating that snapshot as a newly observed action. Tests now cover changed state in reconnect snapshots and repeated identical notifications.

Let me know if this is cleaner.

tests/test_smartbridge.py
alexgorbachev: Agreed. The test bridge now remembers the processor selected during initialization, and accept_connection() no longer takes a processor argument. Reconnects therefore use the same processor instead of defaulting to Caséta. Coverage includes Caséta, RA3, and QSX reconnects.
Thank you for pointing this out.

src/pylutron_caseta/smartbridge.py
alexgorbachev: I’ve kept state processing explicit because it is required, while raw-event subscriptions are optional. This preserves the existing sequence: update device state, notify existing device subscribers, then deliver raw events.

Making state processing another subscriber would require preserving its ordering and existing error behavior through that mechanism. I think this would grow the scope of the change much more. If desired - maybe another refactoring PR?

alexgorbachev:
@mdonoughe  thank you for review. I think addressed all feedback except one where broader refactoring works be required. Let me one what you think.

I've adjusted PR description too.

PR #214, opened 2026-08-26
Add scene programming: read and write virtual button presets

funkzy:
This PR adds the write side of scenes. The library can discover scenes and press them (`activate_scene`), but the scenes themselves have to be created and edited in the Lutron app. With this change an application (e.g. a Home Assistant scene editor) can read what a scene does, edit it, create new scenes, and delete them.

**API** — two methods on `Smartbridge`:

- `get_scene_assignments(scene_id)` — one dict per zone in the scene: level, fade, delay, control type, ids.
- `set_scene(scene_id, assignments, name=None)` — saves the scene: makes its preset match the given list exactly (creating, updating in place, and deleting the typed preset assignment resources as needed), optionally renaming it. An empty list clears the scene; programming an unprogrammed virtual button creates one.

```py
assignments = await bridge.get_scene_assignments("21")
assignments[0]["level"] = 30
await bridge.set_scene("21", assignments)
```

`set_scene` is deliberately declarative ("make the scene exactly this") rather than an add/update/remove family, because the diffing involves bridge behavior that belongs in the library, not in every caller: assignment ids repeat across resource types (matching keys on the full resource href), a zone whose device changed control type needs delete-and-recreate, the bridge enforces one assignment per zone, and each write costs a few hundred milliseconds so unchanged assignments are skipped. If you'd prefer finer-grained assignment operations as the public surface, the internals support layering them — happy to restructure. Also happy to split this into read-only + write PRs if that's easier to review.

**Verified on hardware** (a Caséta Smart Bridge Pro), including live creation, editing, activation, and clearing of a scene controlling a real dimmer. Notable protocol findings baked into the implementation and docs: preset bodies report each assignment twice (typed collections + a legacy untyped `PresetAssignments` list, which is recognized and skipped); `FadeTime`/`DelayTime` on these resources take decimal-second strings, not the `hh:mm:ss` used by the zone command processor; virtual button names cap at 50 characters. Presets holding assignment kinds not supported here (fan speeds, RA3/QSX tuning types) make `set_scene` raise rather than saving a scene that silently keeps rows the caller never saw — those kinds can be added from captured protocol evidence.

Scope of testing: verified against Caséta only; RA3/QSX untested. 19 new tests cover the diff paths (create/update/skip/delete), Switched semantics, control-type change, clear, rename, input validation (parametrized), the legacy collection, and non-zone targets.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `BUTTON_STATUS_LONG_HOLD` constant for the `LongHold` LEAP button event
   emitted by HomeWorks QSX processors.
+- Scene programming: `get_scene_assignments` reads what a scene does, and
+  `set_scene` programs a scene's preset to an exact set of zone assignments
+  (creating, updating, and deleting the typed preset assignment resources),
+  optionally renaming the scene. Programming an unprogrammed virtual button
+  creates a new scene. Verified against a Caséta Smart Bridge Pro.
 
 ## [0.29.0] - 2026-06-09
 
--- a/README.md
+++ b/README.md
@@ -102,6 +102,59 @@ reports direction, command, motion, or source attributes, support can be
 extended from captured protocol evidence without changing the general
 `OpenCloseStop` discovery classification.
 
+### Programming scenes
+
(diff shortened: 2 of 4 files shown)


funkzy:
Thanks for considering this PR. I wanted to see if there are any q's I could answer or anything you'd like changed?

Best,
Jeff

